PR #3, opened 2018-01-17
Added Russia Server

mastad0n:
Added Russia Server: https://dns.d0wn.biz/


jedisct1:
Unfortunately, that server doesn't seem to work.

PR #12, opened 2018-03-01
Add publicarray-au

publicarray:


jedisct1:
Yay!

Thank you!

Other OpenNIC resolvers have a name that starts with `opennic-` but this is not really important.

Added, thanks and welcome to this new resolver!

publicarray:
Oh ok 😄

Actually the resolver was in v1 but was removed some time later because I had broken my cert renew script at some point.

jedisct1:
No problem :) Welcome back!

publicarray:
hm the server is having issues, might be best to remove it again

publicarray:
ok I've seemed to have fixed it for now.

PR #16, opened 2018-03-13
removing ev-ca! sorry to the people using it try using ev-us2.

evilvibes:
I'm removing ev-ca as I don't want to be dealing with both digitalocean and vultr. I'm going to keep everything under vultr as I don't much like digitalocean.


jedisct1:
No problem!

PR #17, opened 2018-03-20
Update public-resolvers.md

xxdesmus:
Updating the stamp for dnscrypt.name


xxdesmus:
yep, this can be rejected. I talked with @jedisct1 

I need to rebuild my server, and I'll send a new PR once ready.

Issue #7, opened 2018-02-08
Adding my server to the public resolvers list (v2) 


EnKrypt:
I'm running a non logging Dnscrypt resolver, and while I've used it on v1 without any issues, I feel confident enough to add it to the public resolvers list now.

However, I've noticed that on v2, the resolver list simply has a `sdns://` URL for each entry. How do I go about obtaining one for my server, and would I need to make any significant modifications on my server/resolver in order to do it?

jedisct1:
Hi EnKrypt!

Your server would be a welcome addition to the list!

Newer versions of dnscrypt-wrapper print the stamp (`sdns://`) but you can also calculate it easily here: https://stamps.dnscrypt.info/

jedisct1:
And if it's running fine, you don't need to change anything to the server.

EnKrypt:
Sounds awesome.

I think I'll update the versions of dnscrypt-proxy and dnscrypt-wrapper that I'm using and make a PR soon.  
Thanks.

jedisct1:
Sounds good!

sergeevabc:
@EnKrypt, 46 days passed, have you changed your mind?

PR #18, opened 2018-03-28
Added my server to the public resolvers list

EnKrypt:
In accordance with DNSCrypt/dnscrypt-resolvers#7


jedisct1:
Thank you! Having a new resolver in India is great! Thanks for running this!

It has been added to the list.

PR #22, opened 2018-04-22
Update for dnscrypt.me

mibere:
I'm sorry, but they delivered me the wrong server. The new and correct one has a 20% higher CPU single thread performance. Now 6 Intel Xeon E5-2620v4 cores.


jedisct1:
No problem :) Merged.

PR #23, opened 2018-05-02
Add publicarray-au2 and publicarray-au-doh

publicarray:
Just trying out GCP with Kubernetes. I still need to workout how well it runs and how much it's going to cost me. If everything works alight, I'll remove publicarray-au soon. Or if it doesn't I'll remove the ones on GCP. 🤞


publicarray:
The configuration is basically a slightly modified (for GCP) version of this: https://github.com/publicarray/dns-resolver-infra

jedisct1:
Added, thanks!

publicarray:
Unfortunately the Google Cloud is more expensive than I thought. Luckily I'm on the free tier but once that runs out I'm going to remove the servers.

PR #25, opened 2018-05-24
removing us2 and us3 servers

evilvibes:
I'm going down to one server as of right now may add more again later. Sorry to the people using the us2 and us3.


jedisct1:
:(

PR #27, opened 2018-05-27
Add new DoH resolver

Belphemur:
Btw, I couldn't add the hashes for the Let's Encrypt certificate. 

I tried with the different Roots of Let's Encrypt  without success.

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -26,7 +26,7 @@ To use that list, add this to the `[sources]` section of your
 DNS-over-HTTPS server running rust-doh with PiHole for Adblocking.
 
 Non-logging, AD-filtering, supports DNSSEC.
-Hosted in Netherland on a RamNode VPS.
+Hosted in Netherlands on a RamNode VPS.
 
 sdns://AgMAAAAAAAAADjE3Ni41Ni4yMzYuMTc1AA5kbnMuYWFmbGFsby5tZQovZG5zLXF1ZXJ5
 
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -22,6 +22,14 @@ To use that list, add this to the `[sources]` section of your
 
 --
 
+## aaflalo-me
+DNS-over-HTTPS server running rust-doh with PiHole for Adblocking.
+
+Non-logging, AD-filtering, supports DNSSEC.
+Hosted in Netherland on a RamNode VPS.
+
+sdns://AgMAAAAAAAAADjE3Ni41Ni4yMzYuMTc1AA5kbnMuYWFmbGFsby5tZQovZG5zLXF1ZXJ5
+
 ## adguard-dns-family
 Adguard DNS with safesearch and adult content blocking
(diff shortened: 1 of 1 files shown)


Belphemur:
Ping ?

publicarray:
@jedisct1 would you mind looking at the PR? looks good to me.

@Belphemur I've never setup a PiHole, so forgive my ignorance but isn't the PiHole logging number of blocked requests as well as top users by IP address and has a full query log? and can this be turned off?

Belphemur:
By default it does, but you can disable it in a single click in the interface.
Moreover, when using DoH all the request are coming from localhost.

I also disable the log for the Nginx that is providing the HTTPS part of DoH.

jedisct1:
I'm awfully sorry for the delay :(

Your resolver has been added.

In order to show the certificates hashes, set the `SHOW_CERTS` environment variable:

```
[2018-06-14 18:58:43] [INFO] [aaflalo-me] TLS version: 303 - Protocol: h2 - Cipher suite: 52392
[2018-06-14 18:58:43] [INFO] Advertised cert: [CN=dns.aaflalo.me] [0e32b8630dd470bc38e665603ea38a8071ed13d6beb893e0237b2ffe08285e14]
[2018-06-14 18:58:43] [INFO] Advertised cert: [CN=Let's Encrypt Authority X3,O=Let's Encrypt,C=US] [3e1a1a0f6c53f3e97a492d57084b5b9807059ee057ab1505876fd83fda3db838]
```

jedisct1:
Thanks again, @Belphemur !

Belphemur:
@jedisct1 Thanks for taking the time to merge it :)

@publicarray Yep it's normal, the basic idea of PiHole is to change the DNS record of the blocked domain to the IP of the server. The server will give back a specific page hosted by the server if accessed using http (will of course fail for https).

jedisct1:
>  the basic idea of PiHole is to change the DNS record of the blocked domain to the IP of the server.

This is how Open DNS made it in order to make money by displaying ads at a time where HTTPS wasn't widely deployed. But I don't think this is a good idea in all other contexts. Especially not in 2018.

This is lying, returning a fake IP. It breaks DNSSEC. Downstream resolvers may cache it. Applications will still consider the lookup as successful (RBL lists...)

And as HTTP (not S) websites are disappearing, users see security warnings that don't make any sense.

DNS defines a `REFUSED` return code. It's been there since the protocol was created.

If accessing something is refused no matter what the reason is, that standard return code should be used instead of pretending nothing happened and return a lie.

Belphemur:
@jedisct1 I completely agree. 

They only added in the latest dev (FTLDNS (their own fork of dnsmasq)) version the possibility to sent a REFUSED answer.

I'm waiting for it to be a little bit more stable (last time I tried, the DNS server crashed) to use it because I agree, this is how it should be.

publicarray:
@Belphemur Thanks for your explanation and the additional server. Yes I also agree with NXDOMAIN answers 👍

PR #28, opened 2018-06-18
qag.me is back

cruisemaniac:
DNSCrypt resolver for Bangalore is back with a new machine and complete opendns upstreaming :-)

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -422,7 +422,7 @@ Maintained by publicarray
 sdns://AgcAAAAAAAAAETM1LjIwMS4yMC4xNzk6NDQzAAtkb2guc2VieS5pbwovZG5zLXF1ZXJ5
 
 ## qag.me
-Plain Vanilla setup of dnscrypt-server-docker on a headless box. NO LOGGING.
+Plain Vanilla setup of dnscrypt-server-docker on a headless box. Upstream to 1.1.1.1.
 Home Server running on a static IP in Bangalore / Bengaluru, INDIA.
 Maintained by Cruisemaniac (https://cruisemaniac.com) aka Ashwin Murali.
 
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -421,6 +421,13 @@ Maintained by publicarray
 
 sdns://AgcAAAAAAAAAETM1LjIwMS4yMC4xNzk6NDQzAAtkb2guc2VieS5pbwovZG5zLXF1ZXJ5
 
+## qag.me
+Plain Vanilla setup of dnscrypt-server-docker on a headless box. NO LOGGING.
+Home Server running on a static IP in Bangalore / Bengaluru, INDIA.
+Maintained by Cruisemaniac (https://cruisemaniac.com) aka Ashwin Murali.
+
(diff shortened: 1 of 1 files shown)


jedisct1:
Yay!

But "no logging" can't really apply if you are forwarding everything to Cisco.

cruisemaniac:
No logging === I do not log. Is there a better way to approach this problem?

jedisct1:
You don't, but they do, and share the data with other companies, and sell access to them.

Hiding the client IP address is not enough for decent privacy. Leaked queries for private host names is already quite sensitive data that shouldn't be recorded by third parties.

If running your own resolver is an issue, maybe choose `1.1.1.1` instead, that has better privacy claims?

ashwinmt:
@jedisct1 `1.1.1.1` is a valid point. Can you point me to running my own resolver? I find it slightly scary to do so. I do need an upstream somewhere? Or do you suggest I handle domain Xfers as well on my own?

ashwinmt:
@jedisct1 updated the upstream to 1.1.1.1 and have also reflected corresponding changes to the text in the PR. Let me know if this makes sense...

publicarray:
> Can you point me to running my own resolver?

I suggest using the docker image, here is a how-to on Scaleway: https://github.com/jedisct1/dnscrypt-proxy/wiki/How-to-setup-your-own-DNSCrypt-server-in-less-than-10-minutes
And one on vultr.com: https://github.com/jedisct1/dnscrypt-proxy/wiki/DNSCrypt-server-with-vultr.com

Shameless plug: (Recommended only if you have a bit of experience with docker). I've implemented a slightly more complex infrastructure with both dnscrypt and doh protocols: https://github.com/publicarray/dns-resolver-infra it is based on the earlier mentioned docker image.

cruisemaniac:
I will do the dnscrypt server in a few days then. When my bandwidth and cpu limitations are much better known. Till then, can this PR be merged so people can use my server as well?

jedisct1:
Merged!

cruisemaniac:
@publicarray I did setup the dnscrypt server. That is what the pull request was about. Maybe I jumped the gun on my response. My question however is to understand if I can setup a server that doesnt log all the way upstream. The DNSCrypt server (docker based - as built by @jedisct1 ) does still need an upstream DNS resolver to handle queries for records it doesnt have.

Who's on the list that I can add to this upstream (`nameserver` entry in `/etc/resolv.conf`) that does not log any queries? According to the PR that @jedisct1 (Thank you my kind sir) has merged, my upstream is `1.1.1.1` (Cloudflare)

What am I missing in this equation? Is there something about the whole DNS circuitry that I've grossly misunderstood? Please help me here.

Thanks,
@cruisemaniac.

jedisct1:
> The DNSCrypt server (docker based - as built by @jedisct1 ) does still need an upstream DNS resolver to handle queries for records it doesnt have.

Not at all.

It resolves everything from the root servers. You don't need any upstream server.

You can install any DNS resolver such as Knot Resolver or Unbound to get the same result.

cruisemaniac:
Okay, wait... My understanding of this setup is pretty flawed then. So, what do I need the /etc/resolv.conf in the docker container for? I need to revisit the source code then! Also, when I setup unbound, what does the forward-addr refer to here?

jedisct1:
The `/etc/resolv.conf` file may be useful for commands used within the container, such as `apt-get install`.

jedisct1:
If you setup Unbound, you can use `forward-addr` to send everything to an upstream server, but you don't need to use `forward-addr`. Just remove the line. That's what that upstream server did.

cruisemaniac:
@jedisct1 noted. Thanks.

PR #29, opened 2018-07-05
Adding dnscrypt.nl-ns0-doh

Yinchie:


jedisct1:
Thank you!

If you are planning to keep using Letsencrypt for a foreseeable future, you can add the Letsencrypt parent certificate hash to the stamp in order to improve security.

Google Chrome and probably other browsers can display certificate hashes, but the easiest way to do it is to start `dnscrypt-proxy` with a `SHOW_CERTS` environment variable:

```sh
env SHOW_CERTS=1 ./dnscrypt-proxy
```

```
[2018-07-05 20:57:33] [INFO] Advertised cert: [CN=itchy.nl] [279b86e2f4a7326580fef60025ba982cc697fb9dd9f5fb2477505d44980c4566]
[2018-07-05 20:57:33] [INFO] Advertised cert: [CN=Let's Encrypt Authority X3,O=Let's Encrypt,C=US] [3e1a1a0f6c53f3e97a492d57084b5b9807059ee057ab1505876fd83fda3db838]
```

So the final stamp becomes:

```
sdns://AgcAAAAAAAAADjEwOC42MS4xOTkuMTcwID4aGg9sU_PpekktVwhLW5gHBZ7gV6sVBYdv2D_aPbg4D2RvaC5kbnNjcnlwdC5ubAovZG5zLXF1ZXJ5
```

If you need to migrate to another cert authority, you can have multiple signatures in the stamp, so you can have both the new cert and the old cert during the migration.

PR #31, opened 2018-07-09
OpenNIC and cryptostorm updates

techmagus:
### I. updated cryptostorm.is sdns and list

- Source: https://github.com/cryptostorm/cstorm_deepDNS/blob/master/dnscrypt-proxy.toml

### II. updated and sync'd OpenNIC list

- Synchronised OpenNIC list between opennic.md and public-resolvers.md.
- Updated ports as declared in https://servers.opennic.org
- Added additional servers in the list
- Made description easier to read
- Removed TumaBox.org from the list. I can not check these servers and the TumaBox.org website has been blocked by ESET NOD32: "The web page is on the list of websites with potentially dangerous content."


jedisct1:
Thanks!

Inconsistencies between `opennic.md` and `public-resolvers.md` should indeed be fixed.

DNSSEC/nologs/nofilters is already present in the stamp, so the description should remain focused on other data.

Cryptostorm servers are not nofilter, as they block ads.

Tumabox works fine, including the website. Don't trust your antivirus except to slow down your computer and increase the attack surface.

PR #32, opened 2018-07-12
Add gridns-sg

sirn:
This is my public DoH server with ads and tracking domain filtering, using [mostly the same block list as Pi-Hole](https://github.com/sirn/gridns-deploy/blob/1418380f8f0a628490f9518fefb74a73194b4ab2/ansible/roles/unbound/files/blocklist/blocklist.txt), although the stack is FreeBSD + unbound. The deployment script (Ansible+Terraform) is published at [sirn/gridns-deploy](https://github.com/sirn/gridns-deploy/) for anyone to audit. I think it will be a useful addition for those of us in South East Asia.

I've been testing for a bit over a week and it seems to work without problems so far.

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -408,6 +408,13 @@ Google DNS (anycast)
 
 sdns://AgUAAAAAAAAAACAe9iTP_15r07rd8_3b_epWVGfjdymdx-5mdRZvMAzBuQ5kbnMuZ29vZ2xlLmNvbQ0vZXhwZXJpbWVudGFs
 
+## gridns-sg
+
+Gridth's public filtering non-logging DNS-over-HTTPS server. Block ads and tracking.
+Hosted in Digital Ocean droplet in SGP region. Upstream to 1.1.1.1.
+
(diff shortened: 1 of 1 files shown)


jedisct1:
Added, thank you!

publicarray:
@jedisct1 maybe we need to add an extra check box to the stamp that is for servers that do the DNS resolution itself and does not forward queries? [[1]](https://github.com/sirn/gridns-deploy/blob/1418380f8f0a628490f9518fefb74a73194b4ab2/ansible/roles/unbound/files/unbound/unbound.conf#L9-L13) [[2]](https://github.com/sirn/gridns-deploy/blob/1418380f8f0a628490f9518fefb74a73194b4ab2/ansible/roles/dnscrypt-proxy/files/dnscrypt-proxy.toml#L2). I am grateful for the servers to take some load off and the addition to blacklists is nice. But if too many servers do this and the server itself is chosen via dnscrypt-proxy (on the server) we end up with a loop in the DNS. In this case there are fallbacks configured to cloudflares DNS so the loop is not much of a problem. Another solution is that we vet servers and add them to a different list. 

What do you think about forwarding servers that forward back into dnscrypt-proxy servers? To me it's just a wired setup and if this continues and we actually end up with a minority actually doing the resolution we could have an issue.

PR #33, opened 2018-07-28
Update public-resolvers.md

ibksturm:
i edited some of my own informations


dimqua:
Shouldn't it be added to opennic resolvers list as well?

jedisct1:
It should, but I can do it.

PR #39, opened 2018-08-21
Update public-resolvers.md

ibksturm:
found german servers of doh-blahdns

added better description o doh-blahdns


publicarray:
@ookangzheng would you like blahdns to be added to dnscrypt-proxy's default resolver list?

jedisct1:
Nice! Thank you!

PR #41, opened 2018-08-22
New DoH resolver

Belphemur:
I've added a proxy to my current PiHole dns server.

This one should be able to reduce greatly the RTT as it will cache the result given by PiHole and also is hosted on Google Cloud Platform.


jedisct1:
Awesome, thank you!

PR #42, opened 2018-08-22
public-resolvers.md add two new resolvers

ianbashford:
Added dnscrypt.uk-ipv4 and dnscrypt.uk-ipv6


jedisct1:
Thank you for setting up these new resolvers!

PR #45, opened 2018-09-16
Add a DoH server from SecureDNS

KrasnayaPloshchad:
The stamp is already published on SecureDNS website for a long times, but I haven't see it in public-resolvers.md, so I decided to add it into this list.


jedisct1:
Thank you!

PR #48, opened 2018-09-22
Add another publicarray server

KrasnayaPloshchad:
The stamp is already published at dns.seby.io, so I paste it to here


jedisct1:
Thank you!

PR #50, opened 2018-10-01
Remove publicarray-au2

publicarray:
I made a mistake the other day I wanted to try and use to google optimised image but I messed up somewhere. I don't have time to fix it now and my free trial is about to run out anyway... It was good while it lasted.

Now that Cloudflare is operating on 1.1.1.1 worldwide I think I can scale back to one server for the Oceana region.

For those interested in how much it would cost monthly on GCP it's $60. To compare Vultr at the time of writing it is $8 a month (1 VPS + 1 Reserved IP) 
 
screen shot 2018-10-01 at 7 32 00 pm


jedisct1:
Sad :(

But yeah, GCP and AWS are really expensive.

PR #52, opened 2018-10-13
Added blahdns jp and de dnscrypt server

ookangzheng:
Added ipv4, ipv6 address


jedisct1:
Added, thank you so much for running these!

PR #54, opened 2018-10-22
Change IP address for aaflalo-me-gcp

Belphemur:
The ip address of my server changed, I forgot to setup a static public one.
This is now done and it shoudln't change in the future.


jedisct1:
No problem, thanks for the update!

PR #75, opened 2019-01-06
Update public-resolvers.md

ibksturm:
did some mistake on ibksturm (dnscrypt server, cert problems)


jedisct1:
No problem, thanks for the quick update!

ibksturm:
hi frank

ok, i will test.. .the interessting thing is, that @home it works and after i uploaded the new keys there was a problem with the signature file... i'll keep you update

ibksturm:
ok i looks as it runs 

grafik

thanks

PR #92, opened 2019-02-07
Fusl's resolvers no longer support DNSCrypt

Zenexer:


Zenexer:
My understanding is that OpenNIC support is also being removed.  doh is being added.

jedisct1:
That's pretty sad :(

Zenexer:
Here's the announcement from @Fusl:

image

Fusl:
Can confirm, please remove, for now at least. Maybe I'll find some time again to fix the bugs that docker and [my auto rollover script](https://github.com/anyqast/uniqast-dnscrypt-wrapper/blob/master/files/dnscrypt-wrapper.sh) introduced but at this time I'm concentrating on getting everything back to more stability.

PR #93, opened 2019-02-19
Going down to one server!

evilvibes:
Now that i have my Canada based server i'll be removing the USA server. It was a good run for that server! Hoping in the future i'll have a west coast Canada server with the east coast server.


jedisct1:
Thanks for having run that server!

PR #96, opened 2019-02-27
Vancouver is here!.

evilvibes:
Added a new server Vancouver and change the Toronto server host. bye bye vultr :(

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -564,6 +564,8 @@ sdns://AQcAAAAAAAAADTIzLjExMS43NC4yMTYgm7NXpUn0viJrTFfNOS7kOq1fgoZXoT6fEkxr8m8Dq
 Non-logging, uncensored DNS resolver provided by evilvibes.com
 Location: Toronto , Canada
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Thanks for the update and welcome to the new server in Vancouver!

publicarray:
Hi @evilvibes I'm currently using vultr. I was wondering why you switched?

evilvibes:
Only move for a host that has both locations I wanted. I also I wanted to be paying in CAD not USD. Vultr is great provider wouldn’t of moved if it wasn’t for them reasons. @publicarray

publicarray:
Thanks @evilvibes, Fair enough, I was just just curious. Keep up the good work 👍

MystesofEternity:
I've been getting incorrect signature warnings from your server since like 2-3 weeks ago. Good thing I just remembered to report about this
image

evilvibes:
@MystesofEternity That's odd everything seems to be working from my end. @jedisct1 What could be causing this issue?

jedisct1:
Maybe a timestamp issue? Can you verify that your clock is correct, @MystesofEternity ?

PR #97, opened 2019-03-02
Add gridns-jp and use new address for gridns-sg

sirn:
I've recently did a [massive update](https://git.sr.ht/~sirn/gridns.xyz/commit/9fb6373708d5514bd4eade498bf6a32c54731ee6) to the `gridns-sg` server, add IPv6 support, and add `gridns-jp` at Tokyo location. This PR updated the existing entry with the new IP address, a new vhost (was: `sg.dns.grid.in.th`, now: `sg.gridns.xyz`), `gridns-jp`, and IPv6 to all `gridns` resolvers.


jedisct1:
Awesome!

Thank you so much for running these!

PR #108, opened 2019-04-07
Update cryptostorm.is servers

df-cryptostorm:
Much needed update to the cryptostorm.is DNSCrypt servers


jedisct1:
Thanks for the update!

PR #115, opened 2019-04-15
add doh.appliedprivacy.net

appliedprivacy:
Hi,

we would like to add our DNS-over-HTTPS resolver to the list.

We didn't fill the IP address and hash field when generating the stamp,
so we are more flexible to point the hostname "doh.appliedprivacy.net" to another IP address should there bee the need or switch to a new certificate authority.

Is the empty hash field supported or does that make it vulnerable to MITM attacks?

(TLS verification should take place via the usual publicly trusted CA)

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -76,6 +76,15 @@ Remove ads and protect your computer from malware
 
 sdns://AQMAAAAAAAAAGVsyYTAwOjVhNjA6OmFkMjowZmZdOjU0NDMggdAC02pMpQxHO3R5ZQ_hLgKzIcthOFYqII5APf3FXpQiMi5kbnNjcnlwdC5kZWZhdWx0Lm5zMi5hZGd1YXJkLmNvbQ
 
+## doh.appliedprivacy.net
+
+Public DoH resolver operated by the Foundation for Applied Privacy (https://appliedprivacy.net).
+Hosted in Vienna, Austria.
+
+Non-logging, non-filtering, supports DNSSEC.
+
+sdns://AgcAAAAAAAAAAAAWZG9oLmFwcGxpZWRwcml2YWN5Lm5ldAYvcXVlcnk
+
 ## arvind-io
 
 Public resolver by EnKrypt (https://arvind.io).


jedisct1:
Hi,

And thanks for running that resolver!

Not providing IP addresses is fine. When IPs are not known in advance, a different resolver is required to resolve the address of the DoH server before it can be used. This adds some latency, but happens automatically.

Hashes are not required, but are important. See https://www.reddit.com/r/dnscrypt/comments/a1asu6/dns_stamp_for_quad9_doh_missing_hashes_compared/eapwxkm/ for why they are.

Hashes of the certificate chain can be printed by defining a `SHOW_CERTS` environment variable and increasing the log level. Unless you are planning to switch to a different certificate authority, you can use the hash of whatever authority signed your final certificate.

jedisct1:
You seem to be using Let's Encrypt, so you can add `3e1a1a0f6c53f3e97a492d57084b5b9807059ee057ab1505876fd83fda3db838` as a hash. This is Let's Encrypt X3 certificate.

This will prevent MITM from other CAs, including root CAs added by Cisco and enterprise surveillance gateways.

appliedprivacy:
thanks for your prompt reply!

> You seem to be using Let's Encrypt, so you can add `3e1a1a0f6c53f3e97a492d57084b5b9807059ee057ab1505876fd83fda3db838` as a hash. This is Let's Encrypt X3 certificate.

To ensure it still works should Let's Encrypt switch their intermediate CA we will add their active and backup CAs [1] in the list of hashes, but I'm wondering why there is a mismatch between the hash you mentioned and the following openssl output:

```
openssl x509 -noout -fingerprint -sha256 -inform pem -in Let\'sEncryptAuthorityX3.crt  
SHA256 Fingerprint=25:84:7D:66:8E:B4:F0:4F:DD:40:B1:2B:6B:07:40:C5:67:DA:7D:02:43:08:EB:6C:2C:96:FE:41:D9:DE:21:8D
```


[1] https://letsencrypt.org/certificates/#intermediate-certificates

jedisct1:
The `openssl` command hashes the whole certificate as-is, whereas what really has to be verified is the trusted section.

https://stackoverflow.com/questions/4803799/how-to-calculate-x-509-certificates-sha-1-fingerprint

This is the `tbs_fingerprint` you can see here: https://censys.io/certificates/9926191203cb4c4ae8170a3263f6cb1d6f997df81afff21587d44d453a89de4a/table

PR #118, opened 2019-04-30
add IPv6 servers

pengelana:
dnscrypt & doh IPv6 servers


jedisct1:
Thank you for adding IPv6 addresses! This is going to be very useful!

Thanks!

PR #119, opened 2019-04-30
add hash to id-gmail-doh-ipv6

pengelana:


jedisct1:
Thanks for adding these.

PR #123, opened 2019-05-06
change commentary, duplicate section name in onion-services.md

4-FLOSS-Free-Libre-Open-Source-Software:
onion-services sections is declared as public-resolvers is commentary only


jedisct1:
Thanks!

PR #124, opened 2019-05-09
OpenDNS IPv4 & IPv6

willowen100:
Added OpenDNS IPv4/IPv6 servers


evilvibes:
The Cisco servers are OpenDNS. Cisco owns OpenDNS.

willowen100:
There are four possible DNS servers available with OpenDNS. The public resolver with DNSCrypt-Proxy only has one listing. That's why I decided to add the spare IPv4 and the IPv6 servers.

jedisct1:
They're all the same, ending up in the same server rack.

PR #127, opened 2019-05-21
removed dns.sb no-logging claim

brainscar:
https://dns.sb/privacy/


publicarray:
According to your policy the clientsubnet is send so that needs to be changed too.

jedisct1:
Thanks for spotting this!

PR #129, opened 2019-05-27
Adds captnemo-in-doh

captn3m0:
- Adds captnemo-in-doh resolver
- Links to Privacy policy for both resolvers (captnemo-in*)
- Confirms DNSSEC/No Logging/No filtering on the DNS Stamp

For the DNS Stamp (DoH Resolver)  I've used the LetsEncrypt X3 certificate in the hash parameter (`25847d668eb4f04fdd40b12b6b0740c567da7d024308eb6c2c96fe41d9de218d` = SHA256 for the LE X3 Certificate). Is that a good idea?

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -115,12 +115,21 @@ sdns://AQcAAAAAAAAAEzE3OC4xMjguMjU1LjI4OjUzNTMgkr1k-Lp2d9IXiFlXoBAgFGZUCJSPW_x81
 
 ## captnemo-in
 
-Server running out of a Digital Ocean droplet in BLR1 region.
+Server running out of a Digital Ocean droplet in Bangalore.
 Maintained by Abhay Rana aka Nemo.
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi,

And thanks for the update!

Unfortunately, it looks like the server is having some issues, and currently returns a 404 error page.

Can you check what's wrong?

captn3m0:
It does resolve: https://doh.captnemo.in/resolve?name=google.com

I'm running [dnss](https://github.com/albertito/dnss) which has a 404 on the home page.

jedisct1:
https://doh.captnemo.in/dns-query?ct=&dns=yv4BAAABAAAAAAABAAACAAEAACkQAAAAgAAAAA

```text
404 page not found
```

jedisct1:
Hi @captn3m0 

Do you think you can fix that issue? 

The server doesn't seem to support DoH. It's a different protocol.

PR #134, opened 2019-06-09
Updated stamp for arvind-io in public resolvers list

EnKrypt:
Updating the stamp post-migration of my dnscrypt resolver


EnKrypt:
Also, a request if it were possible to add my public resolver on the DnsCrypt map.

(https://dnscrypt.info/map)

PR #135, opened 2019-06-10
Add brahma-world

brahma-dev:
- Adds DNSSEC/No Logging/Ad+Trackers+Malware filtering DNS server


jedisct1:
Hi!

And thanks for running this cache!

The certificate hash for Let's Encrypt doesn't seem to be correct. `3e1a1a0f6c53f3e97a492d57084b5b9807059ee057ab1505876fd83fda3db838` is the one I get when trying to connect to your server.

Also, this doesn't seem to be a DNS resolver. Queries are forwarded to Google DNS, so this can't be considered a non-logging server.

brahma-dev:
@jedisct1 Sorry. That was a configuration error. I am using Pihole to filter DNS and it was pointing to Google and Cloudflare instead of the local resolver.

PR #139, opened 2019-07-12
updated sdns stamp for qag.me.

cruisemaniac:


jedisct1:
Hi Ashwin!

Does it do ad blocking?

If it does, the previous stamp looks like it was correct. The only difference between the previous one and your update it the "No filter" flag.

But if your server does ad blocking, that flag shouldn't be set.

cruisemaniac:
@jedisct1 Yes, it does ad blocking. but the stamp is definitely incorrect. I was unable to reach the server using the stamp. The PR changes just the stamp and nothing else.

cruisemaniac:
oh damn! Wait. Please do not merge it. Let me cancel the PR. It looks like I need to change the stamp!

cruisemaniac:
@jedisct1 Fixed the stamp. Now do your magic please :-)

PR #143, opened 2019-07-21
Added Switzerland DoH and Dnscrypt

ookangzheng:
only IPv6


jedisct1:
Thank you for these great additions!

PR #144, opened 2019-07-21
setup server ibksturm new

ibksturm:
did new rawinstall my server, so have to generate new secret.key and public.key


publicarray:
Thanks, FYI you can just copy the files over to the new server. For example with scp.

ibksturm:
> Thanks, FYI you can just copy the files over to the new server. For example with scp.

cool, yar ok, find out how :) cool 
now let's look to STOP the update @jedisct1 @publicarray

publicarray:
It might be too late now but for future reference?

ibksturm:
as i understand its just in time ^^ frank merged some other server... 

and yeah, something new learned 😍

publicarray:
Ah ok I misread the notification. Lucky! Happy to have helped 🤩

ibksturm:
yeah, in german theres a slang / „i-told-you“

can‘t see the forrest ‚cause theres to many trees

PR #145, opened 2019-07-26
Change aaflalo-me-gcp server for new one

Belphemur:
GCP started to be too costly for the traffic I'm getting.
Moving to Ramnode. New IP, new server.

Still providing the old one until the switch is over.


jedisct1:
Thank you for the update!

PR #147, opened 2019-07-26
Re-add aaflalo-me-gcp for backward compatibility

Belphemur:
I realise by removing it, it will cause some people setup to stop working.

This way, it will still work but use the new server.

I only realised this after a night of sleep.


ibksturm:
@Belphemur *lool* sometimes a night in bed makes problems away :)

jedisct1:
Nothing's better than a good night sleep!

PR #151, opened 2019-08-29
Add developer.li opennic dns servers

piraces:
This PR adds two new public DNS Servers located in France & Germany.
Both servers listen on IPv4 and IPv6 addresses, providing DoH and DNSCrypt services.


jedisct1:
Thank you for running these!

Unfortunately, all the IPv6 ones have an invalid stamp.

IPv6 addresses should be specified in the `[ab::cd::ef]:443` form.

Can you update these?

piraces:
Yes of course!

Thanks for noticing the error.

Should be both DoH and DNSCrypt stamps in this format?
Like: [2001:41d0:701:1100::21b5]:8443 (for DoH) and [2001:41d0:701:1100::21b5]:443 (for DNSCrypt)

jedisct1:
Yes, exactly!

PR #156, opened 2019-10-01
Update opennic.md

ibksturm:
new isp :)


jedisct1:
Thanks for the update!

PR #160, opened 2019-10-02
Update blahdns both servers with new crt and key

ookangzheng:
Crt and key expired today, so I regenerated a new one. Enjoy


jedisct1:
Thank you!

PR #161, opened 2019-10-04
add tiarap.org

pengelana:


jedisct1:
Thanks!

Any chance to also have DNSCrypt for these?

With [encrypted DNS server](https://github.com/jedisct1/encrypted-dns-server) it's really easy, and you can even do it on port 443 while still serve DoH.

pengelana:
Thanks!

Yeah, we are thinking to add more services in the future including DNSCrypt with `encrypted DNS server`

PR #162, opened 2019-10-05
Add containerpi doh

hzbd:
- Adds containerpi-doh resolver
- Confirms DNSSEC/No Logging/No filtering on the DNS Stamp


jedisct1:
Thank you!

PR #166, opened 2019-10-17
Update public-resolvers.md

ibksturm:
Ad new Stamp and also new Annon Route :)

Thanks a lot @jedisct1


jedisct1:
Thank you!!!

Too keep things organized, the anon routes are going to be in a new file.

PR #169, opened 2019-10-21
Add anon-publicarray

publicarray:
Hope I'm not too late to join the party!

BTW I have made a copy of the docker image but without unbound. My idea was to closer match the the docker philosophy of one service per container: https://github.com/publicarray/dns-resolver-infra/tree/master/dnscrypt-server. It also makes easier to swap out the DNS backed if people desire.

### DIFF
--- a/v2/relays.md
+++ b/v2/relays.md
@@ -41,3 +41,9 @@ sdns://gSBbMjQwMDo2MTgwOjA6ZDA6OjVmNzM6NDAwMV06MTQ0Mw
 Anonymized DNS relay provided by evilvibes.com Location: Vancouver, Canada
 
 sdns://gQ4xMDQuMzYuMTQ5LjE3Nw
+
+## anon-publicarray
+
+Anonymized DNS relay hosted in Sydney, Australia and maintained by Sebastian Schmidt (@publicarray)
+
+sdns://gRIxMzkuOTkuMjIyLjcyOjg0NDM


jedisct1:
Hi Sebastian,

And welcome to the party!

The new server should also give you much better performance compared to `dnscrypt-wrapper`.

Being able to use an external resolver is a good idea. On the other hand, having a single container that already includes a preconfigured resolver is convenient.

Maybe your container could be a base image, on top of which the version with Unbound is built.
So that both images are available, but share pretty much everything.

publicarray:
Hey!
Thanks, yea it is very convenient to get started that way.

Hm interesting idea, you could overwrite the entrypoint.
So I made some small changes that I could reverse, some where more for personal preference,
* I removed the init scrips (they where no longer needed)
~~* I used Debian buster as a base image~~ Switched back ubuntu
* Changed the binary path to `/usr/local/bin/encrypted-dns` (personal preference but I like that it's in the $PATH)
* Added a dns service discovery option `-d` useful when starting the container and having to wait for the dns resolver (unbound) to be online first and to find the IP of that container)
 * Added a `bin` option to just run the `/usr/local/bin/encrypted-dns` binary (useful to get the version number quickly or custom config file etc. `docker run --rm publicarray/dnscrypt-server bin -h`)

jedisct1:
These are really cool changes!

`encrypted-dns` could be installed using the Debian package, so we don't even need to install Rust.

publicarray:
Thanks!

I kind of like that I can update / compile an update independently of the package maintainers. It’s actually part of the reason I switched my setup to docker. Also I use docker multi-stage builds so it has no impact on the final container image file size.

jedisct1:
Indeed, the multi-stage build is a very good idea.

I'm not too fond of having packages that aren't part of the base system and haven't been installed as packages in `/usr`. But maybe the `PATH` variable can be set in the `Dockerfile`?

publicarray:
Yea good point, I’ll do that once I’m done with work today. 
I’m just wondering if the 2 docker files should be in seperate repos? Or a single repo?

jedisct1:
A single repo is probably fine.

PR #174, opened 2019-10-28
Server does not forward to Google

brahma-dev:
I had pointed this out initial pull request, but somehow got left out. https://github.com/DNSCrypt/dnscrypt-resolvers/pull/135#issuecomment-503516255


jedisct1:
Thank you!

PR #176, opened 2019-11-01
Add cryptostorm's relays

df-cryptostorm:


jedisct1:
Wow, this is amazing!

Thank you for providing relays in so many locations!

jedisct1:
Hi @df-cryptostorm ,

Looks like the Cryptostorm relays have been only configured to relay to port 443.

Which is fine, but quite a few servers run on different ports. For example, Quad9 uses port 5443.

This is the complete set of ports being used so far (same as the list in the `example-encrypted-dns.toml` file):

```
443, 553, 853, 1443, 2053, 4343, 4434, 4443, 5353, 5443, 8443, 15353
```

Do you think you could allow these extra ports to be relayed to?

Thanks!

df-cryptostorm:
Sure... aaand done.

jedisct1:
Wow, that was fast!

Awesome!

PR #177, opened 2019-11-02
securedns.eu updated

ricklahaye:
Updated:
- ads-securedns
- ads-securedns-ipv6
- securedns-ipv6


jedisct1:
Thanks a lot for the update, Rick!

jedisct1:
Hi Rick,

Do you think it would be possible to change the port number you use for the new servers, so that they are compatible with relaying?

This is the set of ports that existing servers use, and that relays allow forwarding to:

```text
443, 553, 853, 1443, 2053, 4343, 4434, 4443, 5353, 5443, 8443, 15353
```

Would one of these work for you?

ricklahaye:
Sorry about that. Will make new merge request.

jedisct1:
No, my bad, I should document this somewhere.

PR #179, opened 2019-11-03
Delete developer.li DNS Servers

piraces:
Developer.li DNS Servers are planned to shutdown (for personal reasons) on November 5, so I open this PR to reflect that in the DNSCrypt repo.

Thank you.


jedisct1:
Thanks for having run these resolvers, Raul!

PR #180, opened 2019-11-03
update opennic-R4SAS stamps

r4sas:
Updating stamps due to loosing previous after unexpected hardware problems.

Post on OpenNIC discuss lists:
https://lists.opennicproject.org/sympa/arc/discuss/2019-11/msg00000.html


jedisct1:
Oops, hardware issues are never fun :(

Thanks for the update!

PR #191, opened 2019-11-29
Update Blahdns Japan IP

ookangzheng:
Thanks


jedisct1:
Thank you for the update!

PR #193, opened 2019-12-02
Update public-resolvers.md

ibksturm:


jedisct1:
Thanks for the update, Andreas!

PR #194, opened 2019-12-03
dnscrypt.one added

mibere:
DNSCrypt server & DNS relay


jedisct1:
Awesome!

Thank you so much for these, @mibere !

PR #195, opened 2019-12-14
Temporarily remove adsnomore servers

jamesponddotco:
Hi 👋🏻,

We are making a few changes to the company, so these servers are being taken down. They but should be back in February or March, but I believe removing them for now is the best option.

They will all be shutdown on **December 31, 2019, at 9 PM UTC**, which I think is enough time for everyone to switch? Let me know if I should leave them for longer, please.

Thanks!


jedisct1:
Hi James,

Sorry for the delayed response, and thanks for the heads up!

This is definitely enough time for people to switch, thanks for the prior notice!

PR #196, opened 2019-12-15
dnscrypt.one, datacenter location

mibere:


jedisct1:
Hi @mibere !

Thanks for the update!

PR #202, opened 2019-12-19
website of dnscrypt.one added

mibere:
My server now has a website


jedisct1:
Beautiful!

Thank you!

PR #204, opened 2019-12-22
update ikbsturm

ibksturm:
some weeks ago, i forgot update the doh-ibksturm and ibksturm stamp in opennic.md file


jedisct1:
Good catch! Thanks!

PR #219, opened 2020-01-23
dnscrypt.one reverted back to a non-filtering server

mibere:


jedisct1:
Thank you!

mibere:
Even if I still think it's, or should be, important to protect users, to be honest, for an operator it's easier to keep things clean and easy. So I removed all filtering from Unbound again.

PR #228, opened 2020-02-15
Updated pi-dns.com servers

NoExitTV:
# PR: Updated pi-dns.com servers
## Summary
1. Updated description of all pi-dns servers to include both city and country.
2. Changed pi-dns WestEU server to new CentralEU server. New DNS-Stamps.

### Reason for PR
 pi-dns have moved the west europe server to central europe. This lead to new ip addresses and as such, new DNS stamps. The "old" west europe server will be online for quite some time before being shut down but it's good to get users to move over asap.


jedisct1:
Thank you for the update! 👍

PR #240, opened 2020-03-19
Updated server details. Previous server was DDOSed to oblivion.

brahma-dev:


jedisct1:
Thanks!

It doesn't seem to be reachable over IPv6 though. Can you check?

brahma-dev:
Really sorry for that. Firewall misconfiguration. Should be working now.

PR #243, opened 2020-03-27
updated static IP for qag.me resolver

cruisemaniac:
New static IP: `49.205.216.229`

DNS Stamp changed accordingly.


jedisct1:
Thanks a lot for the update!

PR #246, opened 2020-03-28
Added relay  anon-sth-se

koki-es:
Updated dnsdist server to 1.5, should accept UDP datagrams larger than 1500 bytes.

## anon-sth-se

Anonymized DNS relay hosted in Sweden.

sdns://gRE0NS4xNTMuMTg3Ljk2OjQ0Mw


jedisct1:
Hi Koki,

Unfortunately, it doesn't seem to work as a relay :(

PR #259, opened 2020-04-12
update cloudflare dns to cloudflare-dns.com

it-can:
Schermafbeelding 2020-04-12 om 10 31 21
Schermafbeelding 2020-04-12 om 11 06 41


it-can:
seems the url was changed... 

https://developers.cloudflare.com/1.1.1.1/dns-over-https/request-structure/

jedisct1:
I wouldn't change that one, which is probably the one cloudflare employees use.

PR #260, opened 2020-04-12
add google-ipv6

it-can:


jedisct1:
The name was already `dns.google`. Looks like you just removed the hash of a certificate :/

it-can:
Is there a good sdns decoder? Got one from npm but it seems dodgy

it-can:
I’ll fix it, will only add ipv6

it-can:
@jedisct1 ok changed the PR so only google-ipv6 is added...

jedisct1:
> Is there a good sdns decoder? Got one from npm but it seems dodgy

You can simply click on the name [here](https://dnscrypt.info/public-servers) :)

Also `dnscrypt-proxy -list -json` will dump the whole list with details as a JSON file.

PR #263, opened 2020-04-17
Added DNSCrypt for SecureDNS.eu & corrected DoH IPv6 Adblock

richieheijmans:
- added both IPv4 and IPv6 DNSCrypt for SecureDNS.eu AdBlock


jedisct1:
Thank you!

PR #264, opened 2020-04-19
SecureDNS removal

ricklahaye:
SecureDNS will be shutting down on the 30th of April 2020.

Thanks everyone for using SecureDNS for the last 3 years.


jedisct1:
Thanks for the great service you provided over these past 3 years, Rick!

PR #267, opened 2020-04-26
Added Blahdns Singapore Dnscryptv2 and DoH server

ookangzheng:


jedisct1:
Thank you! This is going to be very useful for users in Singapore!

PR #274, opened 2020-05-01
update add cloudflare-security-ipv6 add cloudflare-security-ipv6-2 add cloudflare-ipv6-2 update cloudflare to primary ip add cloudflare-2

### LINKED ISSUE #250: cloudflare-security not working
Using dnscrypt-proxy 2.0.19, cloudflare-security is not working.  PARSE_ERROR in the query log.  cloudflare works fine.

4-FLOSS-Free-Libre-Open-Source-Software:
fix #250
update cloudflare-security to primary ip
add cloudflare-security-2
add cloudflare-security-ipv6
add cloudflare-security-ipv6-2

source:
https://developers.cloudflare.com/1.1.1.1/1.1.1.1-for-families/

add cloudflare-ipv6-2
update cloudflare to primary ip
add cloudflare-2

source:
https://developers.cloudflare.com/1.1.1.1/setting-up-1.1.1.1/

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -295,14 +295,14 @@ sdns://AQMAAAAAAAAAEjE4NS4yMjguMTY4Ljk6ODQ0MyC8rDL61UNpFx8IMtYHUCfDIIzu8Ojpn5QY3
 
 Cloudflare DNS (anycast) - aka 1.1.1.1 
 
-sdns://AgcAAAAAAAAABzEuMS4xLjEAEmRucy5jbG91ZGZsYXJlLmNvbQovZG5zLXF1ZXJ5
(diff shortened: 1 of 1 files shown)


jedisct1:
Why?

jedisct1:
`cloudflare-security` uses `1.0.0.2`, and this IP is listed in the page you linked.

Your change adds a port (not required, since `443` is the default) and changes it to `1.1.1.2`, which is also listed on the page. But there shouldn't be any changes of behavior between both IPs.

`1.1.1.0/8` is avoided because some routers are using it as a local network address, so it cannot be reached.

4-FLOSS-Free-Libre-Open-Source-Software:
> `1.1.1.x` is avoided because some routers are using it as a local network address, so it cannot be reached.

ok, thanks for note

But can use IPv6 secondary?

PR #281, opened 2020-05-02
update anon-meganerd

woftor:
- Updated sdns anon-meganerd. Old address will be available for some time to make transistion smooth
- Added anon-meganerd-ipv6


jedisct1:
Thanks for the update!

PR #285, opened 2020-05-12
Add adfree.usableprivacy.net DoH server

markushuber:
* Stamp does not include hash yet, is it possible to pin only the let's encrypt CA instead of the leaf certificate?


jedisct1:
Hi Markus!

Yes, of course, and since Let's Encrypt doesn't support intermediate certificates yet, this isthe recommended approach.

The Let's Encrypt CA hash has been added for you.

Thanks!

PR #286, opened 2020-05-13
Remove gridns-jp and gridns-jp-ipv6

sirn:
Due to day job and personal circumstances, I can no longer maintain a public DNS any longer. I wish to come back at this someday but I have to shutdown the service for the time being :cry:


jedisct1:
No problem, Kridsada.

Thanks a lot for the great service you provided to the community!

PR #293, opened 2020-05-31
Removed qag.me from list

cruisemaniac:
I have a hardware failure which means that I cannot run the dnscrypt resolver anymore until I get my hands on a new kit.

:cry:


jedisct1:
:( Hardware failures are never great.

Thanks for the heads up, Ashwin!

PR #307, opened 2020-06-20
Remove dns.aaflalo.me

Belphemur:
I'm not providing this server anymore.

I don't have the time to maintain the always crashing PiHole that can't deal with 2M request per day.


jedisct1:
I think Pi-Hole was only designed for personal usage. And for that purpose, it is great. But for something publicly exposed, even from a security perspective, it may not be a good idea.

Anyway, thanks for the update!

Belphemur:
On the security side, the dns server was never exposed, only nginx providing DoH and DoT :)

But I admit, I wasn't expecting that much load. I've had to reengineer completely the other server (that I do also use myself) using PowerDNS Recursor instead of PiHole.

ibksturm:
@Belphemur take also a look on dnsdist, i updatet ibksturm.synology.me 3 months ago, cause of exploding DoH requests

runs mutch better than nginx -> dnsproxy -> unbound 

please contact me directly by mail if you have furder questions

PR #309, opened 2020-06-21
update faelix-doh to have DoH and DNSCrypt resolvers

maznu:
This PR has DNS Stamps for all of our open DoH and DNSCrypt resolvers, with IPv4 and IPv6 addressing in both the UK and CH.


jedisct1:
Awesome, thank you Marek!

PR #313, opened 2020-06-26
Update Cloudflare stamps with v3-style multiple IP addresses

hugepants:
Great work on the new `v3` server definition format!


hugepants:
I realised that the 1.1.1.x addresses have compatibility issues, so I have left all the IPv4 stamps alone and updated only the IPv6 stamps (with a minor reorder for readability).

jedisct1:
Thank you!

Yes, some routers are still using 1.1.1.x addresses :(

PR #315, opened 2020-06-28
Quad9 DNS stamps v3 consolidation

hugepants:
Consolidate Quad9 stamps, correct some hostnames, and make format consistent with each other.

Related - https://github.com/DNSCrypt/dnscrypt-proxy/pull/1390

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -1611,116 +1611,68 @@ https://101.101.101.101/index_en.html
 sdns://AgcAAAAAAAAAACAoPxWWFWiOuUdTdn7SvYpzbNqr_iDmmJrktihy4wca5gxkbnMudHduaWMudHcKL2Rucy1xdWVyeQ
 
 
-## quad9-dnscrypt-ip4-filter-alt
+## quad9-dnscrypt-ip4-filter
 
-Quad9 (anycast) dnssec/no-log/filter 149.112.112.9
+Quad9 (anycast) dnssec/no-log/filter 9.9.9.9 / 149.112.112.9
 
+sdns://AQMAAAAAAAAADDkuOS45Ljk6ODQ0MyBnyEe4yHWM0SAkVUO-dWdG3zTfHYTAC4xHA2jfgh2GPhkyLmRuc2NyeXB0LWNlcnQucXVhZDkubmV0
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi!

Quad9 maintain and publish the list themselves: https://www.quad9.net/quad9-resolvers.md

The `public-resolvers.md` file contains an automated copy of it.

Leveraging the v3 format drastically reduces the number of entries here. On the other hand, it means that the same resolvers will have different names according to if you are using them from `public-resolvers.md` or from their list.

I'm not sure what the best thing to do would be here. Less entries is great. So is consistency :/

hugepants:
I see!

The Quad9 list has some errors with the hostnames (e.g. `dns9.quad9.net` instead of `dns10.quad9.net` for the `no-filter` entries). And the hash entries have an unneeded comma. Not that any of these would necessarily cause any issues, but it does highlight the potential problems of relying on an external source. Given that it seems unlikely that the IP addresses would ever change, taking back control of these entries should be relatively maintenance-free.

Regarding the different names, I agree that it's not ideal to have inconsistent names and changing them could break `v3` users who have configured static `server_names`. But that's likely to be a very small minority of users at this early stage, until `v3` becomes more widely used. Which means, if we are going to start changing entry names, it would be best to do it sooner rather than later!

I could add back the original names alongside the new names [in the `broken_implementations` list](https://github.com/DNSCrypt/dnscrypt-proxy/pull/1390).

I do like the idea of having a 'cleaner' `public-resolvers.md` that's potentially a little easier to maintain, but it's not a deal-breaker if we need to keep some 'redundant' entries!

jedisct1:
(just to clarify why: the `v2` list is automatically generated from the `v3` one)

hugepants:
Yes, I see that now - we could minimise the impact by appending `-pri` to all the new names, which would limit the breakage only to those who use static `server_names` consisting of _only_ `-alt` entries. What do you think?

jedisct1:
This is a very good idea!

Having a `-pri` suffix looks a bit odd, but will indeed minimize breakage. There are probably very few people, if any, using only `-alt` entries.

hugepants:
Agree, I'll submit the PR shortly.

PR #316, opened 2020-06-30
Add Safe Surfer DNS servers

BobyMCbobs:
Add Safe Surfer DNS servers


jedisct1:
Welcome to `safesurfer` and thank you for running these servers!

PR #321, opened 2020-07-03
Append -pri suffix to Quad9 entries to minimise breakage

hugepants:
As discussed here - https://github.com/DNSCrypt/dnscrypt-resolvers/pull/315


jedisct1:
Thank you!

PR #336, opened 2020-08-07
changed DoH port

koki-es:
changed in ams-doh-nl, bcn-doh, sth-doh-se


jedisct1:
Thanks for the update!

PR #337, opened 2020-08-12
Updated resolver.

brahma-dev:
Switched to a different server provider.


jedisct1:
Thanks for the update!

jedisct1:
(by the way, only the `v3` file needs to be updated. `v2` files are automatically generated from the `v3` files when they are signed)

PR #348, opened 2020-10-07
Added an ad/tracker/analytics blocking DNSCrypt server from Hyderabad…

coolquasar:
…, India - Oracle Cloud

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -1332,6 +1332,11 @@ Unknown logging policy.
 
 sdns://AgUAAAAAAAAAEFsyMDAxOjQ3MDoyMDo6Ml0gPhoaD2xT8-l6SS1XCEtbmAcFnuBXqxUFh2_YP9o9uDgMb3JkbnMuaGUubmV0Ci9kbnMtcXVlcnk
 
+## DNS Crypt resolver in Hyderabad, India, Oracle Cloud
+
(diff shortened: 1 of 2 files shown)


jedisct1:
And you're in!

The resolver is currently named `jitender` but can be set to something else if you prefer.

Thanks for running that resolver!

coolquasar:
Thankyou so much for approving the PR. I wanted to give back to the
community.

Regards
Jitender

coolquasar:
I just noticed that the sdns string does not match the one I checked in. Is
that an intentional change ? Can I checkin the correct version ?

[image: image.png]

coolquasar:
Alright, I just noticed, that the stamp has been re-generated using updated
public key. Please ignore my email

Regards
Jitender

jedisct1:
Indeed, since this is a filtering resolver, the `no filter` flag has been unset. Also the port number (`:443`) has been removed since this is the default port. That saves a few bytes :)

coolquasar:
Thankyou




----------
On Wed, Oct 7, 2020 at 7:57 PM, Frank Denis <notifications@github.com>
wrote:

Indeed, since this is a filtering resolver, the no filter flag has been
unset. Also the port number (:443) has been removed since this is the
default port. That saves a few bytes :)

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/DNSCrypt/dnscrypt-resolvers/pull/348#issuecomment-704974197>,
or unsubscribe

PR #361, opened 2020-10-29
Update dnscrypt uk

ianbashford:
New host for dnscrypt.uk-ipv{4|6}   and the anon relays.
I'll let requests roll off the old server onto the new one before shutting it down.


ianbashford:
@jedisct1  could you take a look this PR?

jedisct1:
Oops, sorry for the delay Ian!

ianbashford:
NP -- thanks !

PR #364, opened 2020-11-03
Add new servers of Freifunk MUC

awlx:
We switched to a new more powerfull setup.


jedisct1:
Thank you for the update!

PR #367, opened 2020-11-09
Update dnscrypt-de-blahdns-ipv6 stamp

ookangzheng:


jedisct1:
Thank you!

(but please send updates against `/v3`, as `/v2` is generated automatically from it, not the other way round!)

PR #370, opened 2020-11-12
Remove yofiji servers, add official njalla

yofiji:
Remove yofiji dnscrypt servers and add the njalla DoH server.


jedisct1:
No more support for DNSCrypt? :(

yofiji:
Just until the new year - need to migrate to a better spec'd box. Hope to bring another up in adifferent location too.

PR #375, opened 2020-11-28
Add the upcoming dnscrypt.pl as a successor to soltysiak

pysiak:
After 7 years it's time to setup dnscrypt.pl on new infrastructure, cleaner setup, up2date software ... with a proper name.
I'm setting up this service and in near future I'm planning to:
- add 2nd 'anti-malware' service,
- add ipv6 variants for both.

A bit later:
- mark `soltysiak` as legacy and to be gradually retired,
- and when times comes remove it.


jedisct1:
Awesome, thanks!

Damn, 7 years already. `soltysiak` is a legend :)

PR #377, opened 2020-11-29
Added Cloudflare IPv4 address

richieheijmans:
The 1.1.1.1 address wasn't listed for Cloudflare.


jedisct1:
This is intentional, due to 1.1.1.1 still being used by some routers.

PR #385, opened 2020-12-06
Update parental-control.md

w3goodies:
Updated information about 'dnsforfamily'. DNS servers now also do safe search enforcement, and block malware websites.


w3goodies (commented):
Added DoH protocol stamp for 'dnsforfamily'

PR #386, opened 2020-12-06
Update public-resolvers.md

w3goodies:
Updated information about 'dnsforfamily'. DNS servers now also do safe search enforcement, and block malware websites.


w3goodies (commented):
Added DoH protocol stamp for 'dnsforfamily'

jedisct1:
Thanks for the update!

By the way, descriptions can be written on multiple lines. That makes them a little bit more readable :)

w3goodies:
> Thanks for the update!
> 
> By the way, descriptions can be written on multiple lines. That makes them a little bit more readable :)

Sorry, and thanks for adjusting it yourself. I will remember it next time. :)

PR #388, opened 2020-12-08
new: [doh-server] dns.circl.lu added (IPv4 and IPv6)

adulau:
Signed-off-by: Alexandre Dulaunoy <a@foo.be>

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -272,6 +272,22 @@ If you are within India, this might be a nice DNS server to use.
 sdns://AQQAAAAAAAAAEjEzOS41OS40OC4yMjI6NDQzNCAFOt_yxaMpFtga2IpneSwwK6rV0oAyleham9IvhoceEBsyLmRuc2NyeXB0LWNlcnQuY2FwdG5lbW8uaW4
 
 
+## circl-doh
+
+DoH server operated by CIRCL, Computer Incident Response Center Luxembourg.
+
+Hosted in Bettembourg, Luxembourg.
+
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi Alex,

Unfortunately, DoH clients don't seem to be able to connect to these.

Both `dnscrypt-proxy` and `dnslookup` fail with

```text
x509: certificate signed by unknown authority
```

Maybe an intermediate certificate is missing in your TLS configuration?

jedisct1:
SSL Labs also reports that the [certificate chain is incomplete](https://www.ssllabs.com/ssltest/analyze.html?d=dns.circl.lu&s=185.194.94.71&latest).

adulau:
Hi Franck,

My fault, I did a typo in the configuration. Sorry. It should be ok ;-)

Cheers and thanks a lot for your work.

jedisct1:
Looks better now 👍 

Are you planning to stick to DigiCert, so that their cert can be pinned with your entries? (see [operational recommendations for DoH servers](https://github.com/jedisct1/doh-server#operational-recommendations))

adulau:
We might change early next year. So that's why I didn't put a CAA records until now.

jedisct1:
No problem, them!

The set of properties in the stamps don't include DNSSEC, but your servers seem to do DNSSEC validation. Is it intentional?

adulau:
Indeed, we have a load-balancing setup for the DNS and there is still one in the pool not doing DNSSEC validation. That's why I didn't check the DNSSEC part.

jedisct1:
...and... you're in!

Thank you for running these resolvers!

PR #389, opened 2020-12-08
So long Toronto!

evilvibes:
With so many good servers being released I'm going to stick to the west coast for Canada with one server. Sadly I think in 2021 I may just close down that server also.


jedisct1:
:(

PR #391, opened 2020-12-19
Add dnscrypt.pl-guardian

pysiak:
So aside to the vanilla dnscrypt.pl I'm about to launch a filtered service called guardian :-)
It's on the same box and same ip, but port +1.

Would that mean it'd be good to update encrypted-dns.toml to include this port?

allowed_ports = [ 443, 553, 853, 1443, 2053, 4343, 4434, 4443, 5353, 5443, 8443, 15353 ]


jedisct1:
Thank you!

Yes, I guess we should add that new port to the list :)

PR #403, opened 2021-01-10
add dnscrypt.uk doh servers

ianbashford:
+ ipv4 & ipv6 doh servers for dnscrypt.uk
I've moved all the dnscrypt.uk together rather than split them


jedisct1:
Thank you Ian!

PR #405, opened 2021-01-21
Modified arapurayil servers

arapurayil:
Removed arapurayil-dnscrypt, modified title, description and dns stamps for for arapurayil-doh-in


jedisct1:
Thanks for the update!

Sad to not offer DNSCrypt any more :(

PR #408, opened 2021-01-24
Updated 'arapurayil' and 'yepdns'

arapurayil:
added dnscrypt and anonymized relay support, ipv4 & ipv6, for 'arapurayil' and 'yepdns'


jedisct1:
Thank you!

PR #409, opened 2021-01-25
Add new dnscrypt server and anon relay

Zackptg5:
I have all of these on the same server so if that won't pan out in the long haul, let me know


jedisct1:
And you're in!

Welcome, Zack, and thanks for operating that resolver!

Zackptg5:
Awesome thanks! Realized that there weren't any resolvers in the midwest so figured I'd make one :)

PR #410, opened 2021-01-25
add dns.therifleman.name

joker349:


joker349:
Any reason my PR is not being merged?

mibere:
> 24hrs logging

But both [stamps](https://dnscrypt.info/stamps) are set to
 
✅ No Logs

jedisct1:
There are other issues here, and I wanted to take it as an opportunity to finally set up checks on pull requests.

Travis is not an option any more for open source project, so CI rules have to be rewritten anyway.

So, these stamps don't work :)

`dnscrypt-proxy` v2 was written around the time DoH was finalized, so it doesn't support DoT. Version ignores such entries, but older versions will not start if entries include unknown protocols.

The DoH stamp is invalid and is not going to work with any client. The host name shouldn't include `http://`.

Unless you are going to frequently switch to different CAs, it is also important for DoH security to include [certificate hashes](https://github.com/jedisct1/doh-server#dns-stamp-and-certificate-hashes). 

Also, it looks like you are not running your own resolver. Queries are sent to Digital Ocean's standard DNS server. So, maybe the `nolog` should not be set, or at least, this should be mentioned in the description.

joker349:
> So, these stamps don't work :)
> 
> `dnscrypt-proxy` v2 was written around the time DoH was finalized, so it doesn't support DoT. Version ignores such entries, but older versions will not start if entries include unknown protocols.
> 
> The DoH stamp is invalid and is not going to work with any client. The host name shouldn't include `http://`.

I have removed DoT stamp and update DoH stamp by removing 'no logging' check and removing `https` from domain name. Thanks for the suggestions

> Unless you are going to frequently switch to different CAs, it is also important for DoH security to include [certificate hashes](https://github.com/jedisct1/doh-server#dns-stamp-and-certificate-hashes).

By this do you mean TLSA certificate hash that can be calculated [here](https://ssl-tools.net/tlsa-generator)? Also, I use Let's Encrypt certificate so does this mean I have to recalculate the hash every time I renew certificate?

> Also, it looks like you are not running your own resolver. Queries are sent to Digital Ocean's standard DNS server. So, maybe the `nolog` should not be set, or at least, this should be mentioned in the description.

I think I don't invade the privacy of users because the Linode DNS I am using does not know the originating IP. I use this resolver myself and the people at my home. Anyways I have removed that check.

Thanks again for your reply!

PR #422, opened 2021-02-18
Update BlahDNS Japan to latest IP

ookangzheng:
1. Replace BlahDNS Japan region to the latest IP (Linode)
2. Rotate with new DNSCrypt Key


ookangzheng:
Should I also edit v2 markdown file?
I will continuously update Singapore, Germany, Finland, Switzerland server info too.

Thanks

jedisct1:
Hi!

Editing `v3` is enough!

Other ones are automatically generated from it by the signature script.

PR #429, opened 2021-02-23
indonesia-unfilter-dns

rimang:
Indonesia unfilter dns


jedisct1:
Hi,

Unfortunately, these entries [don't appear to work](https://github.com/DNSCrypt/dnscrypt-resolvers/pull/429/checks?check_run_id=1962398404). 

This is a list of DNSCrypt, DoH and Anonymized DNS servers. DoT and other legacy protocols are not included in `dnscrypt-proxy`.

An SNI name is required in order to connect over HTTPS. For security, hashes are also highly recommended. See the [operational recommendations](https://github.com/jedisct1/doh-server#operational-recommendations) in the DoH server documentation.

Feel free to resubmit your resolver! But please check that these entries are valid first :)

jedisct1:
Oops, thanks for the update :)

Just one nits: the port number is not required when it is the default port (`443` for all protocols).

This doesn't seem to be a resolver, but a forwarder to the Google public resolver. This should be mentioned in the description, as it has privacy implications.

PR #434, opened 2021-02-26
Update cryptostorm's resolvers

df-cryptostorm:


jedisct1:
Thank you for the update!

PR #435, opened 2021-02-26
That’s all folks

evilvibes:
It’s been a fun ride thank you for all the hard work Frank with the Project!. 

With everything going on in the world and all the changes happening and what not I just I do not have time to maintain this right now I may return later on.


jedisct1:
Hi Andrew,

And thank you so much for having operated that resolver for all that time. Your contribution is very valuable.

Thanks again, and hopefully see you soon :)

PR #438, opened 2021-03-04
Update anon-ams-nl

koki-es:


jedisct1:
Whoops, that doesn't seem to be valid: https://github.com/DNSCrypt/dnscrypt-resolvers/pull/438/checks?check_run_id=2034825232

jedisct1:
Scratch that, nevermind, this is the CI script that still doesn't handle relays.

PR #440, opened 2021-03-10
add pf-doh

pfactum:
Signed-off-by: Oleksandr Natalenko <oleksandr@natalenko.name>


jedisct1:
I'm still wondering wether "cached via Cloudflare" and "no logging" are compatible.

Maybe we should introduced a new property flag for that kind of resolvers, as well as the ones that are just frontends to other resolvers.

pfactum:
Valid concern, of course. If those flags get implemented, I will consider dropping Cloudflare caching just to be on the safe side.

jedisct1:
But the way, Cloudflare doesn't cache `POST` requests. So if you want caching benefits for DoH, you need to force request to be made over the `GET` method (if you are using [doh-server](https://github.com/jedisct1/doh-server) just add the `-P` flag to the command-line).

PR #443, opened 2021-03-13
merge google

toorich:


jedisct1:
As you probably noticed after testing that localized change, it doesn't work.

Entries with multiple stamps are assumed to share the same properties.

This is not a bad thing. Looking at the list of selected resolvers immediately suggests what variants or protocol a specific server is being used for. Some people also explicitly want to use IPv4 or IPv6 (IPv4 may be more expensive, but that doesn't apply if it is a local server).

Distinct entries are also required for users who still have to use the v1 or v2 formats.

toorich:
> As you probably noticed after testing that localized change, it doesn't work.
> 
> Entries with multiple stamps are assumed to share the same properties.
> 
> This is not a bad thing. Looking at the list of selected resolvers immediately suggests what variants or protocol a specific server is being used for. Some people also explicitly want to use IPv4 or IPv6 (IPv4 may be more expensive, but that doesn't apply if it is a local server).
> 
> Distinct entries are also required for users who still have to use the v1 or v2 formats.

But with the current configuration google or google-ipv6 is not currently available on my side.

PR #446, opened 2021-03-15
v3/relays: add anon-pf

pfactum:
Signed-off-by: Oleksandr Natalenko <oleksandr@natalenko.name>

PR #447, opened 2021-03-15
v3/relays: add anon-pf

pfactum:
Re-submitting #446 after fixing Oracle Cloud firewall. See my investigation in the previous PR.

Please let me know if it works for you now.

FWIW, I've left default `allowed_ports`. Is it encouraged to unrestrict them?


jedisct1:
Works great now!

Thank you for figuring it out!

The default `allowed_ports` is fine, no need to unrestricted them.

Thank you for running this!

PR #450, opened 2021-03-26
Readded meganerd-doh-ipv4

MeganerdNL:
Apparantly the doh-part crashes when cache is full
Measures were taken to prevent this (hopefully)


jedisct1:
This is weird. Do you have more information about how it crashes? Such as a stack trace or something?

jedisct1:
But thanks for the update no matter what. Great to have it back!

MeganerdNL:
> This is weird. Do you have more information about how it crashes? Such as a stack trace or something?

At this moment: no.
I thought is had something to do with cache, but this was not true I found out after some testing.
I use your encrypted-dns-server with doh-proxy

The weird thing was: It was accepting doh on ipv6 and DNSCrypt but it would not reply to doh on ipv4.
It was not fully crashed, so no crash-log.
It started working again after restart of encrypted-dns-server (did not need to restart de doh-proxy).
Both are started as services via systemd on Debian 10.
There is ufw firewall installed on that server and fail2-ban. 

Any ideas how to troubleshoot this are welcome.

Cheers.

PR #452, opened 2021-03-28
add doh to New Jesey server and name change

jlongua:
Thanks for the doh-server update and detailed config explanation, i have now added to New Jersey server.


jedisct1:
Can you check `plan9-ns2-ipv6`? The IPv4 version is fine, but it doesn't seem to be reachable over IPv6.

jlongua:
Thanks, looking into it, let you know when I get it sorted.

PR #458, opened 2021-04-05
New AdGuard DNS resolvers stamps

ameshkov:
Last year we changed the main IP addresses of AdGuard DNS. It took us quite a long time to migrate DNSCrypt to the new addresses. Finally, it's done.


jedisct1:
Thank you for the update, Andrey!

Issue #453, opened 2021-03-29
Add Mullvad servers


yitzhaq:
Would you be willing to add Mullvad's new DoH servers?

https://mullvad.net/en/help/dns-over-https-and-dns-over-tls/

Announcement:

https://mullvad.net/en/blog/2021/3/3/mullvad-doh-and-dot-beta-release/

Cheers

jedisct1:
For a VPN service, it's a little bit sad that they don't support DNSCrypt, that would allow anonymization.

Can you send a pull request with names, descriptions and stamps for the DoH server?

yitzhaq:
I've taken a stab at it. Hope I didn't miss the mark too badly..

jedisct1:
But really, it would be better if you could test  :)

jedisct1:
Merged, thank you!

yitzhaq:
Thanks for the quick merge!
A few marginally related questions:

* As mentioned I used hostname instead of addr. However I noticed that for some entries (such as Cloudflare), 1.0.0.1 was used instead of one.one.one.one. In the case of Cloudflare, that means 1.1.1.1 is seemingly not used at all, which seems suboptimal. Is this intentional, and does the stamps spec support >1 addr entry in a single stamp, as I understand it does for hashes?
* Is there a preference for using addr instead of hostname, or is there another reason for not using eg. one.one.one.one?
* Again looking at the Cloudflare example, one possible benefit could be that this allows for splitting up IPv4 and IPv6 into separate entries. Is this desirable? Like Cloudflare's one.one.one.one, Mullvad's hostnames have both A and AAAA records.
* I also noticed the Cloudflare entry did not appear to have a hash? Is there a particular reason for this, and do you prefer to have (or not have) hashes in the stamps?

Depending on the above answers, I can adjust the Mullvad entries as preferred (if needed).

jedisct1:
A name is always required. This is the name used for SNI, i.e. for the server to know what HTTP service you are accessing.

It is only also used as a host name if no IP address is provided.

If the server has a stable IP address, the IP should always be provided. Without an IP, DoH requires a horrible hack where another server needs to be accessed insecurely ("fallback resolver")  in order to know the address of the actual DNS server.

But if the IP addresses changes constantly, unfortunately, using a fallback server is inevitable.

Hashes are important for security. They were removed for Cloudflare because they are changing their certificates constantly, and without any prior notice.

yitzhaq:
[...]

PR #459, opened 2021-04-06
Added Mullvad resolvers

yitzhaq:
Added entries for Mullvad's resolvers, as requested in #453

Note that:

* This is untested
* This is my first go at generating stamps, based on https://dnscrypt.info/stamps/
* Only hostname is used, not addr
* The SHA256 is the fingerprint of the "Go Daddy Secure Certificate Authority - G2" intermediate currently used to issue the leaf cert(s)
* The stamps do not include bootstrap_ipi, despite these being recommended by https://github.com/jedisct1/doh-server#operational-recommendations, since the online generator didn't seem to support it
* I've made no attempts at updating minisig after my changes
* Mullvad declined to provide a description of their service, so the tagline should merely be considered a suggestion


jedisct1:
Can you remove the DoT entries? This is not supported since there is DoH already.

jedisct1:
The hash doesn't match either according to the CI output.

`dnscrypt-proxy -show-cert` prints it.

yitzhaq:
Thanks for the pointers, hopefully this take two works a bit better. Having now learned how to test, initial tests look OK to me (and checks seem to agree).

PR #460, opened 2021-04-08
add Emerald Onion DoH resolver

natesales:


pfactum:
Are the servers US-based only, so there's no anycast in Europe?

natesales:
Yes, currently US only.

jedisct1:
Hi,

This server doesn't seem to support HTTP/2, which is required by the DoH protocol.

You should also add the hash for the certificates: `3286ff65a65faf32085eea1388c3738ba7e37873c906cce3c4a28b4cc2a58988`

PR #477, opened 2021-04-27
publicarray-au2 signature change

publicarray:
I lost the original while running some docker commands


jedisct1:
Thank you for the update!

PR #481, opened 2021-05-05
dama.no sa-a80

euh2:
Second try, I pretty sure the tests will pass now.


euh2:
I see that I didn't have to create a new PR. I could just have made a new commit to the previous one. Anyway, hope it passes.

PR #487, opened 2021-05-12
update hash id-gmail-doh

pengelana:


jedisct1:
Thanks!

It was already updated to contain the 3 Let's Encrypt certs. But having a shorter stamp isn't bad.

Thanks!

pengelana:
The certificate just renew this afternoon and I thought it still using old hash. Thanks!

PR #492, opened 2021-05-15
Add zackptg5-us-pit resolver/relay

Zackptg5:
I'm moving and sadly found no resolvers with low enough latency for my standards so made a new one!


jedisct1:
Thank you so much for that new resolver!

PR #498, opened 2021-05-25
Add new server Altername

aer0nix:
IPv4-only non-filtering and non-logging DNSCrypt-server with support EmerDNS and OpenNIC. Located  in Moscow, Russia.
Having Beta status due stability testing.


ibksturm:
хорошо 😃 болшои спасибо (а нет я не говарю блоха пар руссий 🤫)

PR #499, opened 2021-05-25
Add Altername

aer0nix:
Fixed pull request [#498](https://github.com/DNSCrypt/dnscrypt-resolvers/pull/498)


jedisct1:
Thank you so much for operating that resolver!

PR #500, opened 2021-05-27
Remove arvind-io from public resolver list.

EnKrypt:
It's been years since I added my dnscrypt resolver to the list here, and I've enjoyed many good years of DNS not being spoofed. 

I've received emails from folks thanking me for making my resolver public, and also emails from folks asking why my resolver was down when I was on vacation.

My ISP no longer does DNS hijacking today. In the end, we won. This means I no longer have a use for my resolver any more. I'm migrating my server and I will not be moving the resolver with it.

It's been a good run, I'm grateful for the battle that dnscrypt fought against censorship, and I'm glad to have been a part of it. It's now time for me to put the sword down.

I hope the day comes when we no longer resolve DNS requests in plaintext and no one has to use this software again. Of course, I say that with all the due love this app deserves.

> So Long, and Thanks for All the Fish


jedisct1:
Hi Arvind,

And thanks a lot for having maintained this public resolver for all that time.
It was very valuable to many other users in India.

And good to hear that some ISPs stopped hijacking DNS, this is great news! Hopefully it will last.

Thanks again, Arvind!

PR #502, opened 2021-05-31
Update Certs  bcn-dnscrypt and sth-dnscrypt-se

koki-es:
Fail rotating logs and forced  to renew certs of these servers


jedisct1:
Thank you for the update!

PR #506, opened 2021-06-07
Update odoh.md

koki-es:
Added  odohrelay-ams  and  odohrelay-bcn


jedisct1:
Already? Wow :)

jedisct1:
`odoh-nl` doesn't seem to work, though. It returns a `Bad Gateway` error.

PR #507, opened 2021-06-07
altername update

aer0nix:


jedisct1:
Hi!

Looks like this would introduce another entry with an existing name.

Is it a new entry or an update to the previous one?

PR #514, opened 2021-06-12
Add UncensoredDNS

livingentity:


jedisct1:
IP addresses haven't changed since 2017, so they may be included in the stamp to avoid a dependency on a bootstrap resolver.

This can also enable having distinct entries for IPv4 and IPv6.

livingentity:
Thanks for merging. :)
Btw is it enough to include just the current certificate hash or what's best practice?

jedisct1:
Adding the current parent certificate (here, Let's Encrypt R3) is fine.

PR #517, opened 2021-06-25
Update public-resolvers.md

jumpsmm7:
Add a dash for consistency.


jumpsmm7:
Thank you for including this @jedisct1

PR #527, opened 2021-07-09
Add openinternet.io DNSCrypt resolver and anonymous DNS relay

dapphp:
Adding a new public DNSCrypt resolver and anonymized DNS relay that I run.


jedisct1:
Hi Drew!

And... you're in!

Thank you so much for running this resolver!

The resolver name is a bit long 😄 This is perfectly fine, but feel free to change it to something shorter if you want!

By the way, if you need external monitoring, [this URL](https://download.dnscrypt.info/status/public-resolvers.txt) has the live status of the resolvers, and [that one](https://download.dnscrypt.info/status/relays.txt) the live status of the relays.

You can use a service such as https://uptimerobot.com to look for `FAIL: <your resolver/relay name>` or the absence of `pass: <your resolver/relay name>` for monitoring.

dapphp:
Hi Frank,

Thanks for merging it in! Agreed on the long name, lol. I think I took the convention from the old spec too literally.

    <provider name> ::= <protocol-major-version> . dnscrypt-cert . <zone> where zone is a valid DNS name.

Oh well. I'll keep that in mind as I add new relays!

Thanks again for all of your work on the protocol and software.

jedisct1:
The provider name is required by the protocol, but in the resolvers list, in can be anything!

It can be changed to `openinternet` or anything you want, without any changes to do server-side (the stamp remains the same).

PR #528, opened 2021-07-10
Added new odoh server

kokial:
Added new odoh server odoh-ams


jedisct1:
Awesome, thank you Alek!

PR #531, opened 2021-07-15
remove current dama.no servers

euh2:
Due to some restructuring those servers will be unavailable soon (in 1-2 weeks). We will provide 1 or 2 new servers which will more stable and maybe faster.


jedisct1:
Thanks for the update! Looking forward to the new servers!

PR #532, opened 2021-07-24
Add new Sydney based servers

gombadi:
Hi. Adding a new Sydney based server I run.


jedisct1:
Hi!

And thanks for running this server! 👍

Looks like the relay doesn't work: https://download.dnscrypt.info/status/relays.txt

Would you mind taking a look?

Thanks again!

PR #534, opened 2021-07-27
Add dns.brahma.world

brahma-dev:


jedisct1:
Welcome back and thanks for the update! 🎉

brahma-dev:
Well, the old server is still active and being used by many, but was removed for some reason. I setup a new instance today and intended to update the IPs, only to find it was removed from the list. Anyways, all is well now. Thanks.

Issue #535, opened 2021-07-28
v.dnscrypt.uk routing


ianbashford:
Hi @jedisct1 

2 questions: 
a)  given the stats below, are you happy to readmit v.dnscrypt.uk
b) There's a question at the end about raising support issues -- any input would be welcomed.

I took a look at https://github.com/DNSCrypt/dnscrypt-resolvers/issues/495  and the routing issue to see if we can safely return v.dnscrypt.uk to the resolver list.

Since 19th July, I've run a lookup every 60 seconds on v.dnscrypt.uk and on your scaleway.fr  for comparison.
For v.dnscrypt.uk, totals are:
   1308 "AU"
   8413 "GB"
   2268 "NL"
    810 "US"

For Scaleway:
   3357 "DE"
   6342 "NL"
   1926 "US"

The routing seems reasonable and comparable for both,  with the exception of the Australian resolutions.
They occurred for a period on 21st -22nd and again on 28th  - mostly its for periods of an hour then they revert, although once the period was a few hours.

I don't think this is a huge issue, so if you're happy with this I'll submit a PR to add the v.dnscrypt.uk resolver back in.

QQ- any ideas how to correctly 'phrase' a question to Vultr support to explore this?

I'm not 100% clear on the path from unbound ->  google's nameservers deciding to return an Australian IP address.  
Is it routing in Vultr, something to do with anycast setup, something else, or something I should ignore :-)     ?
Any input gratefully received.

ianbashford:
Closed via merge https://github.com/DNSCrypt/dnscrypt-resolvers/pull/537

PR #537, opened 2021-07-31
readmit v.dnscrypt.uk and add doh for all dnscrypt.uk servers

ianbashford:
Changes to:
readmit v.dnscrypt.uk as per comments in https://github.com/DNSCrypt/dnscrypt-resolvers/issues/535  
Add doh servers all dnscrypt.uk and v.dnscrypt.uk  [now hosted on separate servers]


jedisct1:
Of course!

Welcome back, Ian!

PR #543, opened 2021-08-14
Set the nolog bit on (additional) faelix dnscrypt entry

taam:
Here's my try, please verify. Thanks!


jedisct1:
Awesome, thank you!

PR #549, opened 2021-08-16
Two new servers

lucenera:
Julia & Rei


jedisct1:
Welcome to these new friends! 🎉🥂

PR #553, opened 2021-08-20
Add openinternet.io ODoH target and DoH resolvers

dapphp:
Thanks to doh-proxy, resolver4.dns.openinternet.io now supports DoH in addition to DNSCrypt and works as an ODoH target.


jedisct1:
Fantastic, thank you!

PR #562, opened 2021-09-07
New relays: anon-curzon-prg, anon-curzon-vie

ztheory:


jedisct1:
Thanks a lot for operating these relays, Zach!

ztheory:
My pleasure, Frank! Hopefully more to come when VPS247 has some available capacity in their other central European PoPs.

PR #567, opened 2021-09-15
adding id-sysadmin-doh server

rimang:
adding id-sysadmin-doh server


jedisct1:
Hi!

So, this is not a resolver. Everything is sent straight to Cloudflare DNS.

Something that cannot be reached using Cloudlfare DNS (ex: the recent `archive.is` issue), will not be reachable either. Instead of a single party that could technically log the data, we now potentially have two :)

Using a proxy to hide the client IP address to a resolver while preserving locality is great! But we now have protocols specifically designed for this, namely Anonymized DNSCrypt and oDoH.

So, instead of a direct proxy to Cloudflare DNS, maybe it would be better to run a DNS relay, or an actual resolver :)

PR #571, opened 2021-09-22
Return of curzon-[vie,prg], add curzon-lux

ztheory:
After running into https://github.com/DNSCrypt/encrypted-dns-server/issues/76 and disabling `daemonize` , these relays ran for ~6 days without the error, so I think they should be here to stay this time.


jedisct1:
Hi!

Looks like `anon-curzon-lux` is having some issues.

Can you take a look?

Thanks!

PR #574, opened 2021-10-04
opennic-R4SAS: update doh pinned certificates

r4sas:
Updating outdated pinned certificates hashes.
And once again: @jedisct1, please add IPv6 records even if your provider can't reach HE.net network. If you can't access it, that not means that nobody can do that.


jedisct1:
After 18 months, the HE IPv6 network just became accessible: https://twitter.com/lafibreinfo/status/1445069021385216001

PR #579, opened 2021-10-29
Update WeVPN servers

Sn0whax:
Added "DNS Firewall" for remaining US, Europe and Asia using toggle "on" here: https://wevpn.com/wedns

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -76,7 +76,7 @@ sdns://AgMAAAAAAAAADzE3Ni4xMDMuMTMwLjEzMCCsFdIhxY-VWoedpSrEKWAhaBEVj-8L-p_FJl6wM
 
 Adguard DNS with safesearch and adult content blocking
 
-Warning: This server is incompatible with anonymization.
+Warning: This server is incompatible with anonymizat
 
 sdns://AQMAAAAAAAAAETk0LjE0MC4xNC4xNTo1NDQzILgxXdexS27jIKRw3C7Wsao5jMnlhvhdRUXWuMm1AFq6ITIuZG5zY3J5cHQuZmFtaWx5Lm5zMS5hZGd1YXJkLmNvbQ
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Still failing: https://github.com/DNSCrypt/dnscrypt-resolvers/runs/4050060938?check_suite_focus=true

v3/public-resolvers.md
```diff
@@ -2177,9 +2177,25 @@ sdns://AQcAAAAAAAAAKFsyMDAxOjE5ZjA6NzQwMjoxNTc0OjU0MDA6MmZmOmZlNjY6MmNmZl0g7Uk9j
...
+
+sdns://AQcAAAAAAAAAFDIzLjIyNi4xMzQuMjQzOjE1MzUzIBf1guPReaoN_V7w-UelcO4YnvqoXnDGsxRPIRc6uSjsHzIuZG5zY3J5cHQtY2VydC5zZWN1cmUuZG5zLnRlc3Q
+
+
+## wevpn-adblock-uk
+
+Private DNSCrypt service by WeVPN in London, UK - https://wevpn.com/dns
+With ad, malware, phishing, cryptojacking blocking (DNS Firewall).
 
-sdns://AQMAAAAAAAAAEjcyLjExLjEzNC45MToxNTM1MyBV2R9ORyXQIRO7dlT0F2mYeEdvXEuBOWMdFkripZcN9B8yLmRuc2NyeXB0LWNlcnQuc2VjdXJlLmRucy50ZXN0
```
jedisct1, line 2182: That was used to work. Why change it?

Sn0whax: Just updating it to follow their sites addresses. If you visit wevpn.com/wedns then scroll down to "anon dnscrypt" then toggle dns firewall as "on" all of the sites (US West, US East, Europe and Asia) start with "sns://AQc" instead of the removed "sns://AQM"

jedisct1: These are all wrong, then. The `No filter` flag shouldn't be set in their "firewall mode", since this is blocking queries.

Sn0whax: I do agree, however most of the ones that wereoriginally added (note, currently listed) are under the "no filter" one. This is why I left the original ones as the "wevpn-country" format then added the "wevpn-adblock-country" for people that do want to use the filtering. wevpn-adblock-useast used to exist as well, however it was renamed to wevpn-useast (not sure of the blocking status of it). While I did initially see some inconsistency, I chose to paste all (yes, even the broken asia/singapore one) into the list so that wevpn users had the options of any country as the previously listed ones didn't really block anything except the wevpn-adblock-uswest.

PR #582, opened 2021-11-05
Sync up with cryptostorm.is/cs-resolvers.md

df-cryptostorm:
derp, haven't pushed updates to this repo in a while


jedisct1:

PR #589, opened 2021-11-17
update sgp-dn53 stamps

poentodewo:
My container error when try to updating with code error exited 111 I was try to restart and still geting error so i decide to rebuild my instance then forgot to backing up the key. the ip still remain the same only public key was changed. since it using azure with tricky to enable ipv6 I will oftern make req when error happen :D


jedisct1:
Thank you for the update!

PR #590, opened 2021-11-17
Update public-resolvers.md

drixter:
Added e-utp.net servers.


drixter:
Hi All,

I've used https://dnscrypt.info/stamps/ page to SDNS and seems to be something wrong, 
Can someone support me? because my lines decode correctly on that page.

Thanks,

jedisct1:
Hi @drixter and sorry for the delayed response. I just moved to a new place and don't have internet connectivity yet.

This list doesn't include DoT and plaintext servers, since these are not supported by dnscrypt-proxy.

So, you should just remove that entry, especially since your service is also accessible over DoH.

Cheers!

PR #592, opened 2021-11-23
Stamps update

bhanupratapys:
Updated ports for dnscrypt to support anonymized dns as most of the relays were using default option for allowed_ports.


bhanupratapys:
Ping @jedisct1, in case you missed this 😅

jedisct1:
Hi! And sorry for the delay. I just moved to a new place and didn't had an internet connection until yesterday.

Thanks for the update!

Issue #599, opened 2021-12-04
dnscrypt.ca: Reinstate IPv6 and Cleanup Stamps


dnscrypt-ca:
Not sure why IPv6 entries were removed.  IPv6 has been relocated to separate hostnames and ports have been included in stamps where possible.  Please replace the existing four dnscrypt.ca entries with these eight entries:

===
## dnscrypt.ca-1

Free, Canadian, uncensored, no-logs, encrypted, and DNSSEC validated DNS service for your pleasure. https://dnscrypt.ca/

sdns://AQcAAAAAAAAAEzE2Ny4xMTQuMjIwLjEyNTo0NDMgGlOjyVB4nL3RCxkzpGibbIRqQPG3PRdSrsrJgp7LfOIdMi5kbnNjcnlwdC1jZXJ0LmRuc2NyeXB0LmNhLTE

## dnscrypt.ca-1-ipv6

Free, Canadian, uncensored, no-logs, encrypted, and DNSSEC validated DNS service for your pleasure. https://dnscrypt.ca/

sdns://AQcAAAAAAAAAKVsyNjA3OjUzMDA6NjE6OTVmOjcyODM6MTFkOTpmODY6ZTY4OV06NDQzICDZGdTsPFAIzccKUeorT8v-ipk8ZDHQ7GgdXqc4RauAIjIuZG5zY3J5cHQtY2VydC5kbnNjcnlwdC5jYS0xLWlwdjY

## dnscrypt.ca-1-doh

Free, Canadian, uncensored, no-logs, encrypted, and DNSSEC validated DNS service for your pleasure. https://dnscrypt.ca/

sdns://AgcAAAAAAAAAEzE2Ny4xMTQuMjIwLjEyNTo0NTMgRE69Z7uD-IB7OSHpOKyReLiCvVCq2xEjHwRM9fCN984UZG5zMS5kbnNjcnlwdC5jYTo0NTMKL2Rucy1xdWVyeQ

## dnscrypt.ca-1-doh-ipv6

Free, Canadian, uncensored, no-logs, encrypted, and DNSSEC validated DNS service for your pleasure. https://dnscrypt.ca/

[...]

PR #600, opened 2021-12-04
Update public-resolvers.md

evilvibes:
Had some free time! Changes for dnscrypt.ca

https://github.com/DNSCrypt/dnscrypt-resolvers/issues/599


dnscrypt-ca:
Thank you very much!  I did try to make the PR myself, but without having local git and Pale Moon spitting out a half-broken interface, it wasn't going well. :-)
Much appreciated. 👍

jedisct1:
Thanks to the two of you!

dnscrypt-ca:
Sorry you guys, I just noticed that the line for "dnscrypt.ca-2-ipv6" is missing the double hashes at the beginning and the stamp for "dnscrypt.ca-2" should be moved up five lines.  I guess it might be easier to delete and paste in the list from #599 rather than deal with the lines individually.

Issue #601, opened 2021-12-14
AhaDNS.com NEW DoH servers and NEW ControlD DoH server provider.


Jefe777:
# AhaDNS.com
You can please add the **NEW DoH servers**:
- [DNS over HTTPS (DoH) – AhaDNS.com](https://ahadns.com/dns-over-https/)
- [DNS Stamps – AhaDNS.com](https://ahadns.com/dns-stamps/)

**Chicago** (DoH)
- IPv4: 193.29.62.196
- IPv6: 2605:4840:3:c4::c4
- Host name: `doh.chi.ahadns.net`

**New York** (DoH)
- IPv4: 185.213.26.187
- IPv6: 2a0d:5600:33:3::3
- Host name: `doh.ny.ahadns.net`

#
# ControlD - Control Your Internet
You can also add to this **NEW DoH server provider**:
- [ControlD - Free DNS](https://controld.com/free-dns/)
- [Generate DNS Stamp | ControlD KB](https://kb.controld.com/en/tutorials/dns-stamp)

## Free resolver servers:
**Unfiltered** (DoH)
- IPv4: 76.76.2.11
- Host name: `freedns.controld.com/p0`
-  Information: No DNS record blocking or manipulation

**Malware** (DoH)
- IPv4: 76.76.2.11
- Host name: `freedns.controld.com/p1`
-  Information: Block Malware

**Ads & Tracking** (DoH)
- IPv4: 76.76.2.11
- Host name: `freedns.controld.com/p2`
-  Information: Block Malware + Ads

**Social** (DoH)
- IPv4: 76.76.2.11
- Host name: `freedns.controld.com/p3`
-  Information: Block Malware + Ads + Social

**Family Friendly** (DoH)
- IPv4: 76.76.2.11
- Host name: `freedns.controld.com/family`
-  Information: Block Malware + Ads + Adult Content + Drugs

**Uncensored** (DoH)
- IPv4: 76.76.2.11
- Host name: `freedns.controld.com/uncensored`
-  Information: Unblock Censored Domains from various countries

#

Thanks.

jedisct1:
Hi!

Can you send a pull request (changing `v3/public-resolvers.md`) for these changes?

Thanks!

Jefe777:
@jedisct1 I do not know how to do it.

jedisct1:
[...]

PR #606, opened 2021-12-20
Add `ahadns-doh-ny`/`ahadns-doh-ny-ipv6` in `v3/public-resolvers.md`

PeterDaveHello:
This is my first PR to add a public resolver here, so I just simply add only one server with both IPv4 & IPv6 stamp.

BTW, I initially found that the dns stamp I generated is much shorted than those in the markdown files, so I decoded other ahadns stamps and found that they're using both Let's Encrypt R3 & Let's Encrypt E1 cert hashes, so I did the same for the consistent, but 
I'm not sure if it's correct, as I only saw the Let's Encrypt R3 cert.

cc #601


PeterDaveHello:
I have no idea why it failed, a little help would be very appreciated.

PeterDaveHello:
Oh, how stupid ... 😅 Thanks!

PR #609, opened 2021-12-20
Delete this request please

Jefe777:
# Delete this request please.

@PeterDaveHello I don't know how to remove/revert the changes so I decided to delete the previous Fork and start from scratch just adding the correct changes.

New Request #611

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -188,6 +188,22 @@ Server statistics can be seen at: https://statistics.ahadns.com/?server=ny
 sdns://AgMAAAAAAAAAE1syYTBkOjU2MDA6MzM6Mzo6M12gzBBg05yDKbYrb7x9DW35MJhpuYHn5jktXNj6QI9NgOYgRE69Z7uD-IB7OSHpOKyReLiCvVCq2xEjHwRM9fCN984RZG9oLm55LmFoYWRucy5uZXQKL2Rucy1xdWVyeQ
 
 
+## ahadns-doh-chi
+
+A zero logging DNS with support for DNS-over-HTTPS (DoH) & DNS-over-TLS (DoT). Blocks ads, malware, trackers, viruses, ransomware, telemetry and more. No persistent logs. DNSSEC. Hosted in Chicago, USA. By https://ahadns.com/
+Server statistics can be seen at: https://statistics.ahadns.com/?server=chi
+
+sdns://AgMAAAAAAAAADTE5My4yOS42Mi4xOTagzBBg05yDKbYrb7x9DW35MJhpuYHn5jktXNj6QI9NgOYgRE69Z7uD-IB7OSHpOKyReLiCvVCq2xEjHwRM9fCN984SZG9oLmNoaS5haGFkbnMubmV0Ci9kbnMtcXVlcnk
+
+
+## ahadns-doh-chi-ipv6
+
+A zero logging DNS with support for DNS-over-HTTPS (DoH) & DNS-over-TLS (DoT). Blocks ads, malware, trackers, viruses, ransomware, telemetry and more. No persistent logs. DNSSEC. Hosted in Chicago, USA. By https://ahadns.com/
+Server statistics can be seen at: https://statistics.ahadns.com/?server=chi
+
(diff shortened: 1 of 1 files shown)


PeterDaveHello:
@Jefe777 thanks, I'm not sure what's the problem to cause the result, the process being used is the same as #601, I'll dig into it and see what can be found.

PeterDaveHello:
@Jefe777 Looks like you're right, I place the query path wrong, thanks for the correction, just if I may, I'll suggest to separate the fix commit with the new added resolvers 😄

v3/public-resolvers.md
```diff
@@ -531,13 +547,22 @@ https://www.comss.ru/page.php?id=7315
 sdns://AQMAAAAAAAAADTk0LjE3Ni4yMzMuOTMgFWnIA4ZtJKvnIs9g6yZT4WIyPb-rQEB27paIxM_OkxMdMi5kbnNjcnlwdC1jZXJ0LmRucy5jb21zcy5vbmU
 
 
+## controld-unfiltered
+
+ControlD Free DNS. Take CONTROL of your Internet. Block unwanted content, bypass geo-restrictions and be more productive. - https://controld.com/free-dns
+
+This is a Unfiltered DNS, no DNS record blocking or manipulation here, if you want to block Malware, Ads & Tracking, or Social Network domains, use the other ControlD DNS configs.
+
+sdns://AgMAAAAAAAAACjc2Ljc2LjIuMTEAFGZyZWVkbnMuY29udHJvbGQuY29tAy9wMA
```
PeterDaveHello, line 556: I guess this one is wrong, the `No filter` option should be checked. Original ` sdns://AgcAAAAAAAAACjc2Ljc2LjIuMTEAFGZyZWVkbnMuY29udHJvbGQuY29tAy9wMA` is correct.

v3/public-resolvers.md
```diff
@@ -573,16 +598,7 @@ ControlD Free DNS. Take CONTROL of your Internet. Block unwanted content, bypass
...
 
-sdns://AgcAAAAAAAAACjc2Ljc2LjIuMTEAFGZyZWVkbnMuY29udHJvbGQuY29tCy91bmNlbnNvcmVk
-
-
-## controld-unfiltered
-
-ControlD Free DNS. Take CONTROL of your Internet. Block unwanted content, bypass geo-restrictions and be more productive. - https://controld.com/free-dns
-
-This is a Unfiltered DNS, no DNS record blocking or manipulation here, if you want to block Malware, Ads & Tracking, or Social Network domains, use the other ControlD DNS configs.
-
```
PeterDaveHello, line 584: I don't think it's a good idea to also change to order of resolvers here.

Jefe777: I think it is better to leave the official order that the dns have on their website:

1. First without filter
2. After with filters
3. From last uncensored.

v3/public-resolvers.md
Jefe777: You have reason, that DNS has no filters.

Jefe777: You have reason, that DNS has no filters.

Do you think **ControlID "Uncensored"** also doesn't have filters?

v3/public-resolvers.md
Jefe777: I think it is better to leave the official order that the dns have on their website: https://controld.com/free-dns

1. First without filter.
2. After with filters.
3. From last "uncensored".

v3/public-resolvers.md
Jefe777: You're right, that dns doesn't have filters the same as ControlID "**Uncensored**", I already corrected both.

v3/public-resolvers.md
Jefe777: I have already moved the DNS as you recommend.

PeterDaveHello (approved):
I'll suggest to separate the fix commit into another PR, otherwise it looks good.

Jefe777:
@PeterDaveHello I do not know how to do that 😣

Jefe777:
@PeterDaveHello I do not know how to do that

PeterDaveHello:
@Jefe777 I believe the commits here tobe merged will be squashed here, so you can simply remove/revert those changes, and send another PR for them if you want, but maybe it is not so important here, just my suggestion.

Jefe777:
Delete please

PeterDaveHello:
@Jefe777 never mind, I might used too technical terms, just mean to make the result here: https://github.com/DNSCrypt/dnscrypt-resolvers/pull/609/files to be very specify to the single PR subject, the remove/revert part can be simply done in the GitHub Web GUI by simply edit the content in the files ;)

Jefe777:
@PeterDaveHello  I'm a newbie, I didn't understand anything, but here is the new Request #611

I don't know how to delete this request.

Jefe777:
@PeterDaveHello  I'm a newbie, but here is the new Request #611

I don't know how to delete this request.

# Delete this request please

PR #621, opened 2022-01-12
Cisco DNS without filtering

HellboyPI:
I found about Cisco OpenDNS without filtering on the wiki:
https://en.wikipedia.org/wiki/OpenDNS

I extracted the ip address from the DOH address:
https://doh.sandbox.opendns.com/dns-query

This ip address is different from the ip address on the wiki page, but it is also valid: https://blogs.cisco.com/security/cisco-interop-discovery-of-designated-resolvers-protocol-implemented

I use this ip address, because the latency is better with this ip address.


jedisct1:
Thank you!

PR #624, opened 2022-01-23
Add dns.digitalsize.net resolver

ol-iver:
This is my own resolver I just added.


jedisct1:
Thank you!

If your server supports ODoH, also feel free to add it to the ODoH servers list!

PR #626, opened 2022-01-30
Update public-resolvers.md

0xb33:


jedisct1:
Thanks!

Looks like this is actually an AdGuard instance, so make sure it is properly secured and configured to receive more than traffic from local networks.

0xb33:
> Thanks!
> 
> Looks like this is actually an AdGuard instance, so make sure it is properly secured and configured to receive more than traffic from local networks.

yes its AdGuard server. i'll add its DOT and DOH too, once i finish my testing!

PR #630, opened 2022-02-10
Update brahma-world IP.

brahma-dev:
Apologies. Didn't know that upgrading AWS EC2 instances changes their IP. Have set an elastic IP now.


jedisct1:
No problem! Thank you for the update!

PR #631, opened 2022-02-12
Update parental-control.md

w3goodies:
Fixed DNSCrypt stamp for DNS for Family.


jedisct1:
Only v3/ needs to be updated :)

PR #640, opened 2022-02-18
New ODoH server in Sweden, odoh-koki-noads-se

kokial:
filter ads and malware.


jedisct1:
Looks like this resolvers is already in the list?

kokial:
No, this is new.
i have review all the "sdns" in *odoh-servers.md* and is different w
another domain (odoh-target-noads-se.alekberg.net)  and unchecked "No
filter"

PR #648, opened 2022-03-12
remove zackptg5-us-il

Zackptg5:
My budget got a lot tighter, unable to pay for hosting anymore :(
Hopefully I'll be able to spin up a new instance in the near future


jedisct1:
Sad to see this server go, but thank you for having run it!
It was certainly useful to many.

PR #656, opened 2022-04-18
correct DNS stamps for PureDNS Singapore

fransallen:
The previous DNS stamps used Indonesian servers, I changed them to Singapore servers which are much more reliable for most cases.

Note: I haven't done a minisign yet. Cc: @jedisct1.


jedisct1:
Thank you!

PR #657, opened 2022-04-20
Add ZynoDNS

itsnebulalol:
ZynoDNS is a fast DNS server based on OpenDNS and Blocky with no logs.


jedisct1:
Thanks for submitting this!

Is it just a forwarder to Cisco DNS?

I'm just a bit confused about the added value. Queries are still logged by upstream resolvers, and clients lose the benefits of anycast, always getting responses from Cisco servers in Ashburn, VA. That doesn't seem fast and adds a new logging opportunity.

Would running your own local resolver be an option instead?

Including forwarders (not anonymizing DNS relays) in these lists can be useful, but they should probably be in their own category.

itsnebulalol:
Hi! Thanks for responding. I was using OpenDNS as the main upstream. I can make it into a resolver from the authoritative nameservers, but it's going to be slower. Should I do that but have OpenDNS/Cloudflare forward too?

PR #667, opened 2022-05-17
updated decloudus DNS stamps due to change in IPs - V3 list

bnkas:


jedisct1:
Thank you for the update!

Only v3 needs to be updated; v2 then gets automatically updated :)

PR #674, opened 2022-05-26
Update parental-control.md

w3goodies:
DNSForFamily now supports DNSSec so I updated the stamps. Also fixed their v6 stamp.


jedisct1:
Thank you!

PR #676, opened 2022-05-29
Remove Gombadi servers

gombadi:
Need to temp remove my servers while I do some hosting rearrangements.


jedisct1:
Thank you for the update!

PR #677, opened 2022-05-31
Update relays.md

kokial:
changed IP

### DIFF
--- a/v3/relays.md
+++ b/v3/relays.md
@@ -30,14 +30,14 @@ sdns://gRZbMjAwMTpiYzg6MTgyNDo3Mzg6OjFd
 
 Anonymized DNS relay hosted in Netherlands - NL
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi!

Looks like `anon-ams-nl` (and the IPv6 version) are not reachable. Can you check?

kokial:
Hi

Now is ready

jedisct1:
It still doesn't seem to relay :)

You can check the relay status here: https://download.dnscrypt.info/dnscrypt-resolvers/status/relays.txt

kokial:
This is weird, i checked everything and seems ipv6 work, but not ipv4.

kokial:
Hi Frank,
I dont find a solution for this issue, maybe something related to packet
fragmentation.
these anon--  are already removed.

BR

PR #683, opened 2022-06-04
Add Surfshark Trust DNS

O7zSO31IVg:


jedisct1:
"Premium secure DNS server" doesn't mean much. Would you mind describe what the server does/supports?

The certificate hash is missing. But the server cannot even be used at the moment since the certificate is expired.

Please test before submitting an entry...

PR #693, opened 2022-06-19
Add dns.digitalsize.net resolver

ol-iver:
My dns resolver has a new home


jedisct1:
Thanks!

If the server supports ODoH, feel free to add it to the `odoh-servers` file as well!

jedisct1:
Looks like the resolver is not running:

```text
{"Status":2,"Comment":"DNS query failure (read udp 172.17.0.2:46360-\u003e172.17.0.1:53: read: connection refused)"}
```

(what weird DoH proxy is returning this?)

ol-iver:
It's up again. There was an error in the startup script resulting in unbound not starting correctly. That's fixed now.
The DoH proxy is is called [dns-over-https](https://github.com/m13253/dns-over-https). Why is the return message weird? The  proxy could not connect to its backend and is telling that.

PR #695, opened 2022-06-21
Corrected init script and timestamp issue

dct-infra:
Corrected init script and timestamp issue


dct-infra:
Several recurring issues that need to be investigated, will fix and get back.

PR #701, opened 2022-06-30
Update public-resolvers.md

jlongua:
add New Jersey, Florida and Mexico resolvers, delete plan9-ns2, plan9-ns2-ipv6


jedisct1:
Awesome!

PR #714, opened 2022-07-28
append saldns resolvers and relays again with new dnsstamps

junkurihara:
Hello!

I would like to put updates and changes of some resolvers and relays hosted in our experimental DNS projects.

- saldns01/02: Due to a serious server failure, they stopped for several days. Now they are finally back. However, I had to regenerate server keys, and hence dnsstamps for servers were updated. (Thanks @jedisct1 for removing invalid dnsstamps! 9a6b8448b800bff1da3a7081ebf44e6ca59f4182)
- saldns03: Due to the  termination of ConoHa service in Singapore region, saldnssg01 has gone (0c1170f8dbca8f3835acfb34d88afeb750b88aeb). So, I set up a new server in Tokyo region.

We already confirmed that they works fine through [another resolver and relay lists](https://github.com/junkurihara/experimental-resolvers). I am happy if this change would be accepted!

Jun


junkurihara:
Any update?

PR #715, opened 2022-07-29
Update plan9-dns Mexico

jlongua:
dnsdist DNSCrypt server has issues, ipv6 can't find useable cert after rotation... I was suprised to find others on this list with resolver certs valid for 1 year and another for a whopping 10 years! So putting encrypted-dns server in front on dnsdist.


jedisct1:
Yay! At least, `encrypted-dns` is going to automatically manage the certificates for you, forever :)

PR #716, opened 2022-08-02
Update plan9-dns New Jersey

jlongua:
I could not get encrypted-dns-server-v0.9.7 to run, had gpf and server crash, tried on Debian 11, Ubuntu 20.04 and 22.04. Encrypted-dns-server-v0.9.6 runs but have memory problems, oom-killer kicks in after a couple hours. What OS and kernel would be recomended?


jedisct1:
I think the issue is in one of the dependencies (`slabigator`). I'm going to look at this as soon as possible.

jlongua:
My pull request seems to stuck on resolverscheck, can you check this. Thanks

jlongua:
will retry, not sure why failed resolvers check

PR #732, opened 2022-09-06
Add bebasdns resolver

lepz0r:


jedisct1:
Is this an actual resolver? It just seems to be a frontend to Cloudflare DNS.

lepz0r:
> Is this an actual resolver? It just seems to be a frontend to Cloudflare DNS.

It forwards to multiple DNS Servers & also block ads & malware, more info here but it's only in Indonesian https://github.com/bebasid/bebasdns

jedisct1:
Thanks for clarifying!

It would be very useful to have this information in the description.

Also, the "no log" flag should probably be unset, as some of these resolvers are logging.

lepz0r:
> Thanks for clarifying!
> 
> It would be very useful to have this information in the description.
> 
> Also, the "no log" flag should probably be unset, as some of these resolvers are logging.

Description is updated & unset the "no log" flag

jedisct1:
It doesn't seem to support DNSSEC nor ECS:

```text
$ dnscrypt-proxy -resolve example.com
Resolving [example.com] using 127.0.0.1 port 53

Resolver      : 202.76.225.74
Lying         : no
DNSSEC        : no, the resolver doesn't support DNSSEC
ECS           : ignored or selective

Canonical name: example.com.

IPv4 addresses: 93.184.216.34
IPv6 addresses: 2606:2800:220:1:248:1893:25c8:1946

Name servers  : a.iana-servers.net., b.iana-servers.net.
DNSSEC signed : no
Mail servers  : 1 mail servers found

HTTPS alias   : -
HTTPS info    : -

Host info     : -
TXT records   : wgyf8z8cgvm2qmxpnbnldrcltvk4xqfn, v=spf1 -all
```

PR #738, opened 2022-09-18
Add new resolver - starrydns

sian1468:
I would like to add my resolver 'starrydns' here.


jedisct1:
Thank you! It's great to have more resolvers in Singapore!

PR #748, opened 2022-11-01
update: blahdns japan dnscrypt

ookangzheng:


jedisct1:
Thanks for the update!

PR #750, opened 2022-11-02
Clean up: Remove dnscrypt.eu from list

simonclausen:
We've gotten to a place where this service is no longer properly maintained. Please merge this to remove the servers from the list in preparation for a shutdown.


jedisct1:
Sad to see it go :(

Thank you for the great service you've been providing to the community for all that time, Simon!

PR #762, opened 2022-11-27
Re-add openinternet with updated stamp

dapphp:
Initially added in PR #527, removed in 1dffcff1 and f9791046.

After some downtime due to hardware failure, the resolver is back up on new hardware using the same IP:port as before. The updated stamp uses a shorter provider name.


jedisct1:
Welcome back, Openinternet!

And thank you for the update!

PR #769, opened 2022-12-24
update alidns cert hash

ZedG2:


jedisct1:
Did it change only in some places?

In Europe I still get the previous cert. Maybe we should keep both?

By the way, your PR changed the DNSSEC/No Filter/No logs -- that shouldn't change :)

PR #770, opened 2023-01-09
Adding my public resolvers in AU

josephbu:
This commit adds 3 public resolvers that I run in Australia, available on both IPv4 and IPv6.

Any questions, let me know!


jedisct1:
Thank you!

If your resolvers support ODoH (they do by default if you're running a recent version of `doh-server`), don't forget to add them to the odoh target list as well.

Thanks!

PR #776, opened 2023-01-14
add dns.sb-doh

Longway22:
DoH URL specified in https://dns.sb/doh/.


jedisct1:
`dns.sb` was removed in 179c9a549d04e50d78217c799ac83508a62d6a4e because of a major bug.

Has this been solved?

Longway22:
@jedisct1 I tried requesting an non-exist domain to DNS.sb server and no 503 but 200 returned, seems the bug has been fixed?

BTW, it doesn't use Cloudflare, according to it's IP address. And it has no logs, according to it says on website.

PR #787, opened 2023-02-09
add la contre-voie DoH server

tamara-schmitz:
The service is described here: https://lacontrevoie.fr/en/services/doh/
Terms and conditions are listed here: https://lacontrevoie.fr/en/cgu/

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -1652,6 +1652,14 @@ Cached via Cloudflare.
 sdns://AgcAAAAAAAAAG1syNjA2OjQ3MDA6MzAzMDo6YWM0MzphZDNiXQANanAudGlhcmFwLm9yZwovZG5zLXF1ZXJ5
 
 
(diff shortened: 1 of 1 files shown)


jedisct1:
Hi!

Is it the same as `42l` ?

`42l` used to be in that list, but folks operating it asked to be removed because that brought more traffic they could handle: 5405f9fb4c82f24cf32f2459efd66b2f778f6a71

Is that solved?

tamara-schmitz:
> Is it the same as `42l` ?
This is indeed the same as 42l. They went through a renaming.

> `42l` used to be in that list, but folks operating it asked to be removed because that brought more traffic they could handle: [5405f9f](https://github.com/DNSCrypt/dnscrypt-resolvers/commit/5405f9fb4c82f24cf32f2459efd66b2f778f6a71)
> 
> Is that solved?

Sorry. I do not know that. This PR is not on their behalf. I'll send them a message.

PR #789, opened 2023-02-11
add easymosdns-doh

UjuiUjuMandan:
As described in the file.


jedisct1:
Hi!

Since this is not a resolver, but a forwarder that relays queries to untrusted 3rd parties, would it be possible to mention this in the description?

Thanks!

jedisct1:
Hashes are also missing; this is important for DoH security. See https://github.com/DNSCrypt/doh-server#dns-stamp-and-certificate-hashes

UjuiUjuMandan:
I've added upstream information and hashes.

UjuiUjuMandan:
Looks like the action failed, probably because the CDN used by this server doesn't support HTTP/2? 

I confirm the hashed I entered is correct.

jedisct1:
Wait, this is also using Cloudflare as a front-end. That should be mentioned as well. That's a lot of 3rd parties, not sure that the "no log" flag applies here. 

The hash should be removed after all, sorry for the confusion. Cloudflare certs are not the same everywhere, so they can't be pinned.

UjuiUjuMandan:
I didn't know it uses Cloudflare before... I sshed to a oversea VPS and the dig result shows Cloudflare IPs. But here in China it shows a different, local CDN (China Mobile).

```
doh.apad.pro.		120	IN	CNAME	dfb10000.pchome.net.cname.hcnamecdn.com.
dfb10000.pchome.net.cname.hcnamecdn.com. 30 IN A 112.47.52.134
dfb10000.pchome.net.cname.hcnamecdn.com. 30 IN A 117.149.203.53
```

Cloudflare firewall may log queries sometimes, so I will remove the nologs flag.

PR #793, opened 2023-03-06
Update public-resolvers.md

c586f85042652:


jedisct1:
The resolver is not reachable.

Please, before submitting it for addition to the public list, make sure to at least use it yourself to ensure that it actually works.

PR #805, opened 2023-04-11
Re-add dns.digitale-gesellschaft.ch

c5aba924-b91a-4504-b8e7-9dd81a704703:
Re-add `dns.digitale-gesellschaft.ch` and combine stamps because the virtual IP addresses (VRRP) are shared by the server instances anyhow.


jedisct1:
Looks good now, thank you!

PR #818, opened 2023-07-16
altername offline

aer0nix:
The server is temporarily out of service.


jedisct1:
Thanks for the heads up! Hope it will be back soon!

By the way, only the file in `v3` need to be updated. Other files are generated automatically when signed.

PR #824, opened 2023-08-12
add tuna-doh-ipv6

UjuiUjuMandan:
They also provide IPv6 access, but unlike 101.6.6.6, it doesn't have a for-IP TLS certificate.


jedisct1:
Thank you!

PR #827, opened 2023-08-23
Update Mullvad DNS servers

iyanmv:
Addresses have changed (doh -> dns) as well as some new blocking levels have been added recently since early this month.

More info here: https://mullvad.net/en/help/dns-over-https-and-dns-over-tls/


jedisct1:
Maybe the previous names shouldn't be changed, so that people who have them in their configuration don't need to change them.

PR #828, opened 2023-08-24
RethinkDNS

HellboyPI:
Regarding the correct certificate hash, the -show-certs command-line flag shows this output for rethinkdns:
[NOTICE] Advertised cert: [CN=rethinkdns.com] [6bd089f78d2efbc7d0a6870a2abe7d7be9245e035831ffa41e2429396d9e5488]
[NOTICE] Advertised cert: [CN=E1,O=Let's Encrypt,C=US] [cc1060d39c8329b62b6fbc7d0d6df9309869b981e7e6392d5cd8fa408f4d80e6]
[NOTICE] Advertised cert: [CN=ISRG Root X2,O=Internet Security Research Group,C=US] [5aef843ffcf2ec7055f504a162f229f8391c370ff3a6163d2db3f3d604d622be]
[NOTICE] Advertised cert: [CN=ISRG Root X1,O=Internet Security Research Group,C=US] [11b102e6b1f63e528984d6025f32b138241fc88bbd7519574d70c9832d53e1e8]

I have used the last hash to create the dns stamp.

### DIFF
--- a/v2/public-resolvers.md
+++ b/v2/public-resolvers.md
@@ -936,6 +936,14 @@ https://www.dns0.eu/
 sdns://AgMAAAAAAAAAACCaOjT3J965vKUQA9nOnDn48n3ZxSQpAcK6saROY1oCGQxraWRzLmRuczAuZXUKL2Rucy1xdWVyeQ
 
 
+## dns0-unfiltered
+
+The unfiltered version of dns0.eu.
+https://open.dns0.eu/
+
+sdns://AgcAAAAAAAAAACCaOjT3J965vKUQA9nOnDn48n3ZxSQpAcK6saROY1oCGQxvcGVuLmRuczAuZXUKL2Rucy1xdWVyeQ
+
+
 ## dnscry.pt-amsterdam-ipv4
 
 DNSCry.pt Amsterdam - no filter, no logs, DNSSEC support (IPv4 server)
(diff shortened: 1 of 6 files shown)


jedisct1:
Hi,

And sorry for the delay.

Can you clarify what " does so in a way that doesn't leak the identity of the client (user)" really mean?

It doesn't seem to support ODoH, so certainly Cloudflare has access to everything?

HellboyPI:
This info is from their FAQ page https://rethinkdns.com/faq under "Is RethinkDNS a recursive resolver?" :

_"Rethink DNS is both a stub and a recursive resolver-- sky.rethinkdns.com forwards user requests to another recursive resolver like Cloudflare's 1.1.1.1, but does so in a way that doesn't leak the identity of the client (user). Think of it like a proxy resolver that sits in-between a recursive resolver and you, the user."_

Also, there is additional info from their Privacy Policy page https://rethinkdns.com/privacy under "DNS Privacy Policy" : 

_"We do not collect, sell, license, share any personally identifiable data from our DNS resolvers. The code is open source so you may inspect it yourself or deploy to your servers. Rethink DNS resolvers do not log requests."_


Personally, even though RethinkDNS has good intentions, Cloudflare has access anyway.

ignoramous:
*rdns dev here*

> does so in a way that doesn't leak the identity of the client (user)

This is more of Rethink's stub resolver at `sky.rethinkdns.com` acting as a "proxy" between Google and Cloudflare public DNSes and the user. It only forwards the contents of the DNS query after stripping it of ECS and the incoming HTTP headers too.

> It doesn't seem to support ODoH

Working a JS implementation for ODoH based on [dajiaji/hpke-js](https://github.com/dajiaji/hpke-js)... some day, it'll be ready and we'll deploy it to Rethink.

ignoramous:
Merged here: https://github.com/DNSCrypt/dnscrypt-resolvers/commit/2320fff8a9e46cfb95a62fb244ab241118d7a287

PR #829, opened 2023-08-29
split digitale gesellschaft

c3d1c06c-bf26-477e-b0eb-c50ef4477ba6:
Split dns stamps, users want both servers available at the same time, e.g. for load balancing or potential downtime.


jedisct1:
A long time ago, "users" asked to group multiple stamps of the same resolver into a common entry in order to make the list more readable and easier to maintain. So, this is what was done.

Applications can do whatever they think is best from these files.

In `dnscrypt-proxy`, a random resolver that is up and running is chosen at startup time, and new resolvers are regularly chosen, after every refresh of the servers parameters. So, load balancing is already there. Client traffic is spread across the available options.

In case a server with multiple entries has entries that suddenly become unresponsive, other servers will be used, and at the next refresh, entries from the original server will be reevaluated, and the ones that are not down will be used.
Maybe an option to try other stamps from a resolver before the refresh can be added. But the current behavior mitigates unwanted traffic spikes from a single client, and people running abusive benchmarks, degrading the service for other users.

In either case, what to do is expected to be handled at application level. This is not that lists' responsibility.

It also doesn't make sense to treat "digitale gesellschaft" as a special case, doing things differently from all other servers. There are other servers with multiple stamps, and we should remain consistent.

PR #835, opened 2023-09-15
Add ODoh-target meganerd

MeganerdNL:


MeganerdNL:
Let me know if it doesn't work. Can't test ODoH unfortunately.

jedisct1:
HI!

Why did you close this?

I didn't reply immediately because it was my girlfriend's birthday.

But shouldn't that be merged?

MeganerdNL:
No it was not because of you. :-). I couldn't make it work and decided to do some more research first. oDoH doesn't seem to work only by adding -O. There is apparently some more configured required, but I don't know what. Maybe you could point me in the right direction.

PR #844, opened 2023-09-23
Update meganerd-doh

MeganerdNL:
Now with the right hash in the sdns address (posted issue earlier today).


MeganerdNL:
I will make a new PR for dnscrypt too.

PR #852, opened 2023-10-27
Add new anon relay in Tokyo

gombadi:


jedisct1:
Fantastic! Thank you so much!

PR #856, opened 2023-11-10
Update odoh-relays.md

kokial:


kokial:
Fixed NAT64 config issue

Issue #807, opened 2023-04-25
FDN DoT/DoH Servers


hucste:
Since 04/19/2023, the FDN — French Data Network — an association to promote Open Internet and Usenet, published new DoT and DoH services! 
_(this is too an ISP)_

⇒ On port 853 for DoT:
IPv4 Adresses: 
- 80.67.169.12
- 80.67.169.40

IPv6 Adresses:
- 2001:910:800::12
- 2001:910:800::40

⇒ for DoH: 
- https://ns0.fdn.fr/dns-query
- https://ns1.fdn.fr/dns-query

See more informations (in French): https://www.fdn.fr/ouverture-des-services-dot-doh/

jedisct1:
Do you mind submitting a pull request with a proper description and stamps?

hucste:
I dont known how to do this!?

hucste:
> Do you mind submitting a pull request with a proper description and stamps?

If you explain me where put the proper description and how to obtain stamps, it could certainly be possible ;)

jedisct1:
The only files to update are:

`v3/public-resolvers.md`

and if your server also accepts ODoH, `v3/odoh-servers.md` as well.

Other versions of the list will be automatically updated from the `v3` one.

The DNS proxy should print the stamps when it starts, or somewhere in the log files.
If not, you can use the [online calculator](https://dnscrypt.info/stamps/)

Also see the [operational recommendations for DoH servers](https://github.com/dnscrypt/doh-server#operational-recommendations) and [how to get certificate hashes](https://github.com/dnscrypt/doh-server#dns-stamp-and-certificate-hashes).

PR #862, opened 2023-12-06
add FDN public-resolvers

GannonTdW:
Hello,

Add fdn public resolver DoT and DoH (mentioned in https://github.com/DNSCrypt/dnscrypt-resolvers/issues/807)

Don't hesitate to tell me any mistakes.


jedisct1:
Are you affiliated with them?

These servers used to be in the list, but they asked to be removed:

> Please remove our public resolver from the list
>
> We've been suffering from an *heavy* traffic from dnscrypt-proxy users. Our little infrastructure isn't ready to have its place on this list.
> 
> We'd be glad if you can remove our proxy from it.
> 
> Thanks for your understanding.

That was circa 2019. Have things changed?

jedisct1:
Ah, my bad, that was another French provider called "42l.fr".

jedisct1:
FDN resolvers have been added.

PR #870, opened 2024-01-15
adding tirapan.top doh servers

demarcush:


jedisct1:
I can't connect to it either when `http3 = true` is set.

PR #874, opened 2024-01-23
Add ns2.ksol.io public resolver

karolyi:


jedisct1:
... and it's fast and works perfectly well.

Thank you for operating that resolver!

karolyi:
Cheers and keep up the good work, let's keep the internet decentralized. :)

PR #875, opened 2024-01-28
dct-at shut down

dct-infra:
The server will remain active for a few days post removal.


jedisct1:
Thank you for having operated that resolver and for the heads up!

PR #886, opened 2024-03-24
Update odoh-servers.md

kokial:
Now targets and relays share the same domain


jedisct1:
Thanks for the update!

PR #890, opened 2024-04-09
updated fluffycat-fr-01

skynet2077:
Apologies for the inconvenience, I accidentally deleted the state file for fluffycat-fr-01 during an update and thus had to regenerate it. They are now backed up for both servers, so it won't happen again :)


jedisct1:
No problem, that happens to all of us!

PR #894, opened 2024-05-05
Re-adding tirapan.top

demarcush:
Problem with caddy and h3:
https://github.com/caddyserver/caddy/issues/5086#issuecomment-2094529691

Problem with DoH proxy:
https://github.com/m13253/dns-over-https?tab=readme-ov-file#known-issues

So unless anyone can suggest a way to simultaneously have the two of them (https server and DoH proxy) listen on 443, all I can do is wait for them to get fixed/implemented.


jedisct1:
Not sure what the compatibility issues are with the DoH server you linked. Maybe try [doh-server](https://github.com/DNSCrypt/doh-server)?

jedisct1:
HTTP/2 is fine. HTTP/3 is blocked in many places.

jedisct1:
QUIC and HTTP/3 are finally coming to the Go standard library: https://pkg.go.dev/golang.org/x/net/quic

This is going to make it way easier for software like Caddy to support these.

jedisct1:
@demarcush All you need is make sure that queries for `/.well-known/odohconfigs` are also routed to the DoH server.

You can then test with the stamp `sdns://BQcAAAAAAAAAC3RpcmFwYW4udG9wEC9ub21lbi1xdWFlc2l0dW0` and an ODoH relay.

PR #896, opened 2024-05-06
Remove tirapan from opennic source

demarcush:
tirapan is no longer an opennic resolver.


jedisct1:
The file is automatically created from the `opennic.subset` list.

Removed, thank you!

PR #897, opened 2024-05-06
Update Adguard

demarcush:
Updated Adguard, including:
- Replacing IP addresses as VHOSTs, because Adguard has IP SAN for its IPv4 addresses
- Adding both reachable IPs for every resolver type.
- Sorting ~~and renaming~~


jedisct1:
Do these entries need to be renamed?

This is going to cause a disruption for people currently using adguard.

PR #900, opened 2024-05-12
More IP SANs, More IPv6

demarcush:
Be sure to take your time as I'm submitting this as a draft first, it's a huge PR.

The goal is to improve the reachablity of servers with using SANs and IPv6, and also correcting mistakes here and then.

Why using IP SANs? Anti-censorship reasons; State actors around the world has started blocking DoH traffic with SNI inspection, removing that single point of failure in the chain results in such actors resort to active probing and/or blocklisting IP ranges, both of which are costly enough to reduce the incentive for them to do as such.


jedisct1:
prcheck found errors: https://github.com/DNSCrypt/dnscrypt-resolvers/actions/runs/9056401552/job/24878890291?pr=900#step:3:916

But yes, splitting this into multiple PRs doing only one class of changes at a time would be much better.

PR #901, opened 2024-05-12
Add NIC.CZ

demarcush:


jedisct1:
Did you test it?

It used to be there, but [was removed in 2022](https://github.com/DNSCrypt/dnscrypt-resolvers/commit/8869baf0fb5ec4d1205a6bdd4027271e07cee2d0) due to broken responses.

jedisct1:
Thank you!

But shouldn't the name be `nic.cz` rather than `cz.nic`?

PR #902, opened 2024-05-14
Update Cisco's OpenDNS

demarcush:


jedisct1:
Cisco DNS has become unusable at least in France and Portugal. The servers return fake responses to all queries.

The problem is that they still return valid certificates even in these countries.

This is very annoying. Not sure what to do with that.

jedisct1:
Yes. It's Cisco themselves returning poisoned responses.

jedisct1:
This is intentional.

```sh
$ dig google.com @208.67.220.220
;; ADDITIONAL SECTION:
google.com.		0	IN	TXT	"Due to a court order in France issued under Article L.333-10 of the French Sport code the OpenDNS service is not currently available to users in France and certain French territories."
```

There may be other reasons, since EU and unwanted data collection don't play well together.

PR #904, opened 2024-05-14
Update DNSPod

demarcush:


jedisct1:
Looks like we don't get the same certs :(

With these entries, all I get is `Certificate hash [8f4b73997c4b04ea50feafa91a2431d42bca6b54c23bcf9dbbf572249394e10c] not found`

Unfortunately, we may have to remove cert hashes here :(

jedisct1:
By the way, do you know of a reliable way to check for GFW filtering? Some specific domains that are always blocked?

PR #905, opened 2024-05-14
Update Cloudflare and Google

demarcush:


jedisct1:
Cloudflare certificates change all the time, and this is also inconsistent depending on the client location. Unfortunately, we can't include hashes for them, this breaks too frequently.

jedisct1:
We should keep `sdns://AgcAAAAAAAAABzEuMC4wLjEAEmRucy5jbG91ZGZsYXJlLmNvbQovZG5zLXF1ZXJ5`.
This is a special entry for `dnscrypt-proxy`, and there's some history behind it, that goes back to before Cloudflare DNS was publicly announced.

jedisct1:
https://www.reddit.com/r/dnscrypt/comments/1dxke0m/dnscrypt_stopped_working_on_multiple_docker/

Crap. Wondering if this is related.

jedisct1:
Yeah, they had a Frankenstein configuration. IP addresses should indeed be fine even for v2 users.

PR #906, opened 2024-05-14
Update Yandex

demarcush:


jedisct1:
IPv6 should remain distinct entries. There's no way to select IPv6 resolvers otherwise.

jedisct1:
I don't think `ipv6_servers` and `ipv4_servers` work with DoH. Will have to check, though; I never use DoH.

jedisct1:
This is the original naming scheme. Some people are still running `dnscrypt-proxy` version 1 or other software that don't even have IPv4/IPv6 selectors.

We can certainly make the current version of `dnscrypt-proxy` smarter regarding IPv4/IPv6 filters, and define a version 4 of the resolvers list. But it will take years before people upgrade.

But to be honest, I like the fact that IPv6 entries are distinct, and that when we use them, this is explicit. For the same resolver, performance can be very different when reached over IPv4 and IPv6, so it's also nice to have both being probed, and used as needed depending on their latency.

jedisct1:
BTW I'm at the SYCL conference, so I won't have much time to review/merge these PRs this week.

jedisct1:
The `-family` variants don't seem to block anything? At least I tried the most popular adult website, and nothing got blocked.

Do they need something else? Maybe a different query path? Do they work at all?

jedisct1:
I would remove it, as it doesn't behave as expected. And not include it in the regular `yandex` since it may still block stuff.

PR #907, opened 2024-05-14
Merge Tirapan entries

demarcush:


jedisct1:
Why?

People may want to explicitly use IPv6.

Also, renaming entries should be avoided as much as possible. It's going to break connectivity for all the people having that resolver in the `server_names` list.

PR #913, opened 2024-05-17
Normalize Adguard and DNS.SB

demarcush:


jedisct1:
The IP address is expected to be in the IP address field.

Not sure what applications will do if they only see an IP address in the host name field. Presumably try to resolve it using a resolver, which is going to fail.

jedisct1:
If it works, this is by accident 😊 That behavior is not part of the specification.

jedisct1:
Because I didn't pay attention to this in your previous PRs :)

PR #917, opened 2024-05-18
Update Quad9

demarcush:


jedisct1:
The Quad9 list is automatically generated from https://quad9.net/dnscrypt/quad9-resolvers.md

jedisct1:
But since Quad9 is also accessible using DNSCrypt (which connects to IPs directly), is it important?

jedisct1:
The script to create the quad9 file is here: https://github.com/DNSCrypt/dnscrypt-resolvers/blob/master/utils/quad9-merge.py

It can be improved to add IP addresses to DoH servers, if they don't change.

PR #922, opened 2024-06-16
Remove Gombadi servers

gombadi:
Sorry may as well remove my servers as they keep getting removed by the testing system as can't be reached. I have not made any changes so it seems to be something outside my control.


jedisct1:
😔

PR #923, opened 2024-06-18
changed techsaviours.org to digitalprivacy.diy

Dan-Sun:
We've changed the name to DIGITAL PRIVACY .DIY instead of TECH SAVIOURS .ORG.  
Therefore, we have set up a new server, new domains and new services.

techsaviours.org dnscrypt will be still maintained and available until 25 Oct 24.


jedisct1:
Hi!

The IPv6 stamps should be different.

How are you running the server?

If you're running `encrypted-dns` directly, in the config file, you can list all the external IP addresses.

For example:

```toml
listen_addrs = [
    { local = "0.0.0.0:443",    external = "65.21.228.41:443" },
    { local = "[::]:443",       external = "[2a01:4f9:6b:1adb::2]:443" }
]
```

Both stamps should then be displayed.

You can also go to https://dnscrypt.info/stamps/ , enter the current stamp and change the IP to get the new one.

jedisct1:
New name is very cool, btw :)

PR #932, opened 2024-07-07
Add `DNSFilter`

demarcush:


jedisct1:
@vomw I don't understand what this link shows.

Is there a link somewhere explaining this resolver's policy?

PR #934, opened 2024-07-08
Normalize `switch`

demarcush:


jedisct1:
I know this is redundant, but we should keep the IP address field here.

jedisct1:
My concern is that without an IP address in the IP field, an implementation may ask a resolver to resolve the host name (or here, an IP address), or assume that only entries with an explicit IP address can be used without a bootstrap resolver.

The overhead is minimal, so I'd be very cautious about potentially breaking changes, especially since this behavior is not documented.

jedisct1:
As an example, the code that generates the resolvers map on `dnscrypt.info` starts with:

```python
for i, resolver in enumerate(data):
    if 'addrs' not in resolver:
        continue
```

Resolvers without an explicit address are ignored.

It's reasonable to argue that it should look at the protocol, and resolve the names, or extract the IP from the host name if it's an IP. But that's just one example of real-world code outside `dnscrypt-proxy` that removing IP addresses would affect.

PR #935, opened 2024-07-12
Update ksol.io public resolver

karolyi:
For some reason, upon updating the docker container `jedisct1/dnscrypt-server`, it changed my server's keys.

This update uses the new keys.


karolyi:
Using the [stamp calculator](https://dnscrypt.info/stamps/), it seems that the only change here is the addition of the port 443 to both IPv4 and IPv6 addresses. I didn't change anything, the public key is the same, so the keys didn't change.

It seems that the update brought this.

jedisct1:
Hi László!

`:443` is optional. So the stamps are effectively the same!

I just have a script that removes `:443` before publishing updates.

The keys are still the same. So everything's good!

Issue #868, opened 2024-01-13
Some dnscry.pt servers are missing in the public-resolvers.md


Wabuo:
For example, the Auckland one ..

@Brueggus is this somehow intended?
Also cheers for hosting the only relay in NZ

jedisct1:
The list is synchronized with the `dnscrypt.pt` source manually. I should really look at automating this.

Is there a file that contains the relays? I randomly tried `https://www.dnscry.pt/relays.md` but there's nothing there.

Wabuo:
Ther's is this 
https://github.com/DNSCrypt/dnscrypt-resolvers/blob/master/v3/relays.md

And this
https://www.dnscry.pt/anon-relays.md

Also go on to his side and scroll down a bit to see all the things he provides.
https://www.dnscry.pt

Brueggus:
> @Brueggus is this somehow intended?

No, I am happy if those are added to the "offical" lists :)

There's also https://www.dnscry.pt/resolvers.json if that helps to keep the resolver lists in sync.

jedisct1:
The relays have been added already. Thanks for the URL!

PR #945, opened 2024-08-12
dnscry.pt update script

TBBle:
Based on #868 and particularly https://github.com/DNSCrypt/dnscrypt-resolvers/issues/868#issuecomment-1911704288, this is a trivial script to update the dnscry.pt entries in `v3/public-resolvers.md` and `v3/relays.md` from https://www.dnscry.pt/resolvers.json.

<details>

<summary> No-longer relevant issues </summary>

The below issues appeared in the first version of this PR, but @Brueggus (dnscry.pt maintainer) has updated the published resolver data to resolve them both, per https://github.com/DNSCrypt/dnscrypt-resolvers/pull/945#issuecomment-2334304634

----

The first run generates a lot of churn, as compared to the data in this repo, the upstream data now encodes the port number in all of its DNS Stamps. It's easy to see this after format.py is run and the `v1/dnscrypt-resolvers.csv` is updated. It would be possible to strip the port 443 from the DNS Stamps and recalculate them, but doing so would make these stamps incomparable to upstream, and that cost me a bit of time trying to work out why they were _all_ different before I checked with https://dnscrypt.info/stamps/.

Also visible in `v1/dnscrypt-resolvers.csv` is some hard-to-avoid churn in entries where upstream has multiple resolvers with the same "location" key, as my dumb-trivial mechanism to resolve that may not pick the same one to rename as the one done previously.

This is visible for example with current `dnscry.pt-amsterdam-ipv4` which ends up as `dnscry.pt-amsterdam02-ipv4` (based on the public key), and the _other_ entry with that location is now `dnscry.pt-amsterdam-ipv4`.

I'm not sure if this is really an issue for end users. An alternative would be to add the suffix to the location for any host with a hostname that isn't `___01`, but that would rename existing `dnscry.pt-singapore-ipv4` (again matching the host keys) to `dnscry.pt-singapore03-ipv4`. In this approach, Tokyo would also ends up with only `dnscry.pt-tokyo02-ipv4` and `dnscry.pt-tokyo03-ipv4`, but that's actually fair since current `dnscry.pt-tokyo-ipv4` is gone, and gets replaced in-place with what upstream calls `tyo03.dnscry.pt`.

I didn't commit the result of running the scripts since I'm not set up for minisig: I'm not sure what the expectation here is, but it would be trivial to add such a commit if desired.

It might also make for cleaner history if I in-place updated all the dnscry.pt DNS Stamps in-place first to have the :443 port number, then the churn will be much more readable.

The current output from execution looks like this:
```console
> py -3 .\utils\update-dnscry.pt-entries.py

[v3/public-resolvers.md]
Duplicate entry: [dnscry.pt-amsterdam-ipv4] => [dnscry.pt-amsterdam02-ipv4]
Duplicate entry: [dnscry.pt-amsterdam-ipv6] => [dnscry.pt-amsterdam02-ipv6]
Duplicate entry: [dnscry.pt-hongkong-ipv4] => [dnscry.pt-hongkong02-ipv4]
Duplicate entry: [dnscry.pt-hongkong-ipv6] => [dnscry.pt-hongkong02-ipv6]
Duplicate entry: [dnscry.pt-losangeles-ipv4] => [dnscry.pt-losangeles02-ipv4]
Duplicate entry: [dnscry.pt-losangeles-ipv6] => [dnscry.pt-losangeles02-ipv6]

[v3/relays.md]
Duplicate entry: [dnscry.pt-anon-amsterdam-ipv4] => [dnscry.pt-anon-amsterdam02-ipv4]
Duplicate entry: [dnscry.pt-anon-amsterdam-ipv6] => [dnscry.pt-anon-amsterdam02-ipv6]
Duplicate entry: [dnscry.pt-anon-hongkong-ipv4] => [dnscry.pt-anon-hongkong02-ipv4]
Duplicate entry: [dnscry.pt-anon-hongkong-ipv6] => [dnscry.pt-anon-hongkong02-ipv6]
Duplicate entry: [dnscry.pt-anon-losangeles-ipv4] => [dnscry.pt-anon-losangeles02-ipv4]
Duplicate entry: [dnscry.pt-anon-losangeles-ipv6] => [dnscry.pt-anon-losangeles02-ipv6]
```

----

</details>

### DIFF
--- a/utils/update-dnscry.pt-entries.py
+++ b/utils/update-dnscry.pt-entries.py
@@ -1,0 +1,189 @@
+#! /usr/bin/env python3
+
+import json
+import os
+import sys
+import unicodedata
+import urllib.request
+
+CURRENT_DIR = "v3"
+
+
+class Entry:
+    name = None
+    description = None
+    stamps = None
+
+    def __init__(self, name, description, stamps):
+        self.name = name
+        self.description = description
+        self.stamps = stamps
+
+    @staticmethod
+    def parse(raw_entry):
+        description = ""
+        stamps = []
+        lines = raw_entry.strip().splitlines()
+        if len(lines) < 2:
+            return None
+        name = lines[0].strip()
+        previous_was_blank = False
+        for line in lines[1:]:
+            line = line.strip()
+            if previous_was_blank is True and line == "":
+                continue
+            previous_was_blank = False
+            if line.startswith("sdns://"):
+                stamps.append(line)
+            else:
+                description = description + line + "\n"
+
+        description = description.strip()
+        if len(name) < 2 or len(description) < 10 or len(stamps) < 1:
+            return None
+
+        return Entry(name, description, stamps)
+
+    def format(self):
+        out = "## " + self.name + "\n\n"
+        out = out + self.description + "\n\n"
+        for stamp in self.stamps:
+            out = out + stamp + "\n"
+
+        return out
+
+
+class DNSCryDotPTEntry:
+    raw = None
+    name = None
+    description = None
+
+    def __init__(self, raw, name, description):
+        self.raw = raw
+        self.name = name
+        self.description = description
+
+    @staticmethod
+    def parse(raw_entry):
+        name = (
+            raw_entry["location"]
+            .lower()
+            .replace(" ", "")
+            .replace("-", "")
+            .replace("'", "")
+        )
+        # The hand-written entries were ASCII-only, so this is done to reduce churn.
+        name = unicodedata.normalize("NFKD", name).encode("ASCII", "ignore").decode()
+        # NOTE: Just trusting this is true for all dnscry.pt resolvers/relays, rather than trying
+        # to process the stamp to confirm this.
+        description = f"DNSCry.pt {raw_entry['location']} - DNSCrypt, no filter, no logs, DNSSEC support"
+        return DNSCryDotPTEntry(raw_entry, name, description)
+
+    def get_ipv4_resolver_entry(self):
+        return Entry(
+            f"dnscry.pt-{self.name}-ipv4",
+            f"{self.description} (IPv4 server)\n\nhttps://www.dnscry.pt",
+            [self.raw["ipv4_stamp"]],
+        )
+
+    def get_ipv6_resolver_entry(self):
+        return Entry(
+            f"dnscry.pt-{self.name}-ipv6",
+            f"{self.description} (IPv6 server)\n\nhttps://www.dnscry.pt",
+            [self.raw["ipv6_stamp"]],
+        )
+
+    def get_ipv4_relay_entry(self):
+        return Entry(
+            f"dnscry.pt-anon-{self.name}-ipv4",
+            f"{self.description} (IPv4 server)\n\nhttps://www.dnscry.pt",
+            [self.raw["anon_ipv4_stamp"]],
+        )
+
+    def get_ipv6_relay_entry(self):
+        return Entry(
+            f"dnscry.pt-anon-{self.name}-ipv6",
+            f"{self.description} (IPv6 server)\n\nhttps://www.dnscry.pt",
+            [self.raw["anon_ipv6_stamp"]],
+        )
+
+
+def process(md_path, resolvers, is_resolvers):
+    print("\n[" + md_path + "]")
+    entries = {}
+    previous_content = ""
+    out = ""
+
+    with open(md_path, encoding="utf8") as f:
+        previous_content = f.read()
+        c = previous_content.split("\n## ")
+        out = out + c[0].strip() + "\n\n"
+        raw_entries = c[1:]
+        for i in range(0, len(raw_entries)):
+            entry = Entry.parse(raw_entries[i])
+            if not entry:
+                print("Invalid entry: [" + raw_entries[i] + "]", file=sys.stderr)
+                continue
+            if entry.name in entries:
+                print("Duplicate entry: [" + entry.name + "]", file=sys.stderr)
+            if entry.name.startswith("dnscry.pt"):
+                continue
+            entries[entry.name] = entry
+
+    for resolver in resolvers:
+        if is_resolvers:
+            v4 = resolver.get_ipv4_resolver_entry()
+            v6 = resolver.get_ipv6_resolver_entry()
+        else:
+            v4 = resolver.get_ipv4_relay_entry()
+            v6 = resolver.get_ipv6_relay_entry()
+
+        # Some entries in the dnscry.pt database used to have repeated location names.
+        # Right now, all entries are unique, so this code shouldn't be needed, but
+        # it'll catch any future oversights rather than silently losing entries.
+        if v4.name in entries:
+            v4name = v4.name
+            v4suffix = resolver.raw["host"].split(".")[0][3:]
+            v4name = v4name.replace(resolver.name, f"{resolver.name}{v4suffix}")
+            print(
+                "Duplicate entry: [" + v4.name + "] => [" + v4name + "]",
+                file=sys.stderr,
+            )
+            v4.name = v4name
+        entries[v4.name] = v4
+
+        if v6.name in entries:
+            v6name = v6.name
+            v6suffix = resolver.raw["host"].split(".")[0][3:]
+            v6name = v6name.replace(resolver.name, f"{resolver.name}{v6suffix}")
+            print(
+                "Duplicate entry: [" + v6.name + "] => [" + v6name + "]",
+                file=sys.stderr,
+            )
+            v6.name = v6name
+        entries[v6.name] = v6
+
+    for name in sorted(entries.keys()):
+        entry = entries[name]
+        out = out + "\n" + entry.format() + "\n"
+
+    if out == previous_content:
+        print("No changes")
+    else:
+        with open(md_path + ".tmp", "wt", encoding="utf8") as f:
+            f.write(out)
+        os.unlink(md_path)
+        os.rename(md_path + ".tmp", md_path)
+
+
+with urllib.request.urlopen("https://www.dnscry.pt/resolvers.json") as response:
+    resolverdata = json.load(response)
+
+resolvers = []
+for data in resolverdata:
+    resolvers.append(DNSCryDotPTEntry.parse(data))
+
+resolvers.sort(key=lambda r: r.name)
+
+process(f"{CURRENT_DIR}/public-resolvers.md", resolvers, is_resolvers=True)
+process(f"{CURRENT_DIR}/relays.md", resolvers, is_resolvers=False)
--- a/utils/format.py
+++ b/utils/format.py
@@ -197,6 +197,7 @@ If you want to contribute changes to a resolvers list, only edit files from the 
     else:
         with open(md_path + ".tmp", "wt") as f:
             f.write(out)
+        os.unlink(md_path)
         os.rename(md_path + ".tmp", md_path)
 
     # Legacy
@@ -214,7 +215,8 @@ If you want to contribute changes to a resolvers list, only edit files from the 
         else:
             with open(md_legacy_path + ".tmp", "wt") as f:
                 f.write(out_legacy)
-                os.rename(md_legacy_path + ".tmp", md_legacy_path)
+            os.unlink(md_legacy_path)
+            os.rename(md_legacy_path + ".tmp", md_legacy_path)
 
     # Historic
 
--- a/utils/format.py
+++ b/utils/format.py
@@ -182,7 +182,7 @@ If you want to contribute changes to a resolvers list, only edit files from the 
     for name in sorted(entries.keys()):
         entry = entries[name]
         out = out + "\n" + entry.format() + "\n"
-        if not name in INCOMPATIBLE_WITH_LEGACY_VERSIONS:
+        if name not in INCOMPATIBLE_WITH_LEGACY_VERSIONS:
             out_legacy = out_legacy + "\n" + entry.format_legacy() + "\n"
 
     if os.path.basename(md_path) == "public-resolvers.md":
--- a/utils/format.py
+++ b/utils/format.py
@@ -244,15 +244,17 @@ If you want to contribute changes to a resolvers list, only edit files from the 
 
     # Signatures
 
-    for path in [md_path, md_legacy_path, csv_historic_path]:
-        try:
-            subprocess.run(
-                ["minisign", "-V", "-P", MINISIGN_PK, "-m", path], check=True
-            )
-        except subprocess.CalledProcessError:
-            signatures_to_update.append(path)
+    if signatures_to_update is not None:
+        for path in [md_path, md_legacy_path, csv_historic_path]:
+            try:
+                subprocess.run(
+                    ["minisign", "-V", "-P", MINISIGN_PK, "-m", path], check=True
+                )
+            except subprocess.CalledProcessError:
+                signatures_to_update.append(path)
 
 
+# Change to None to skip signature updates, e.g. during development.
 signatures_to_update = []
 
 for md_path in glob(CURRENT_DIR + "/*.md"):
--- a/utils/subset.py
+++ b/utils/subset.py
@@ -1,8 +1,6 @@
 #! /usr/bin/env python3
 
-from copyreg import constructor
 import sys
-from glob import glob
 
 
 class Entry:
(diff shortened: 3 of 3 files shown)


utils/update-dnscry.pt-entries.py
```diff
@@ -0,0 +1,190 @@
...
+    @staticmethod
+    def parse(raw_entry):
+        name = (
+            raw_entry["location"]
+            .lower()
+            .replace(" ", "")
+            .replace("-", "")
+            .replace("'", "")
+        )
+        # TODO: The hand-written entries were ASCII-only, so this is done to reduce churn.
```
TBBle, line 75: Ooops, this isn't a TODO anymore, as I actually did it. But maybe worth discussing naming generally, given the duplicate-name issue when using (mangled) location which isn't unique upstream.

Brueggus:
dnscry.pt maintainer here. Thanks for your great work! Keeping this repository in sync with the resolver lists I publish has been haunting me for months, so I am happy to see the progress made in this PR!

> The first run generates a lot of churn, as compared to the data in this repo, the upstream data now encodes the port number in all of its DNS Stamps.

That's (unfortunately) due to laziness: I am taking the DNS Stamps for DNSCrypt and Anonymized DNS straight from the output of encrypted-dns-server. When I started the project, I only supported DNSCrypt, so there was no need to calculate any other DNS Stamps. Now changing this wouldn't be a big deal since I already calculate the stamps for DoT/DoH. I'll look into that.

> Also visible in v1/dnscrypt-resolvers.csv is some hard-to-avoid churn in entries where upstream has multiple resolvers with the same "location" key, as my dumb-trivial mechanism to resolve that may not pick the same one to rename as the one done previously.

I have never properly implemented having multiple resolvers in the same location and that's why things become inconsistent. At the moment, I am (ab-)using the `location` field, which you find in the JSON as well, and add an incrementing value if needed. For example, the resolver `tokyo-ipv4` in my `resolvers.md` (https://www.dnscry.pt/resolvers.md) shows `"location": "Tokyo"` in the JSON, `tokyo02-ipv4` shows `"location": "Tokyo 02"` and so on.

I will have to make some adjustments here, but I don't think there's a better/more proper way than adding an incrementing value to locations which host more than one resolver.

Besides that: Is there anything I can change in the JSON output to make things easier for you?

TBBle:
I think the stamp data in the JSON is fine, my concern was merely that _this repo_'s existing (hand-maintained) stamps are either old, or were being recalculated to remove the port before publication, and so servers that haven't actually changed are showing a changed stamp when you compare the generated output to the existing data. That's why I was leaning towards a second PR to go first, which would update all the DSN stamps for dnscry.pt entries to include the port number, matching upstream, but not otherwise changing the content. That makes the diff resulting from running this PR's script much smaller.

As far as Location, then yeah, making them unique by including an incrementing key or something (I'd think the same as the DNS name, ideally) to ensure stability as servers appear and disappear makes sense. Tokyo is an example of where it's working well already, the concern was about Amsterdam, Hong Kong, and Los Angeles, where they do not have any such integer, and my quick-workaround mismatched the existing data in one of those cases, possibly surprising users if they update and don't re-select the desired service. Having the name be unique and upstream defined would let me remove the hack, and then any such mismatch or churn will only happen once, and be isolated to places where this repo and your upstream data source have an existing mismatch.

If there wasn't existing data, I wouldn't use Location as the unique name anyway, I'd prefer to derive it from the host-name. That would introduce a once-off churn for all downstream consumers of this list who are using dnscry.pt servers though, so it's probably not feasible at this point.

----

Of course, as you are the dnscry.pt maintainer, you are welcome to take this script, run it manually, apply any manual fixups or churn reduction you see fit, and submit the results as a PR against the data files. That might make things easier for the repo owner, as they would more-easily trust a data-only PR from you compared to a script from (random passing stranger) me.

Brueggus:
>  or were being recalculated to remove the port before publication

I think that's the case. IIRC the stamps were taken as they are first and the port has been removed in a later commit.

I have just published new resolver lists (+ JSON) which have the port removed if the default port is used, which is the case for all resolvers at the moment. This change was overdue anyway to be compliant with the official (?) [specifications](https://dnscrypt.info/stamps-specifications/#dnscrypt-stamps) for DNS stamps.

> the concern was about Amsterdam, Hong Kong, and Los Angeles, where they do not have any such integer

Oh boy... until now I didn't even notice those were missing the incrementing key. I wonder how the clients handled the resolver lists containing two resolvers with the same identifier. Anyways, as a quick fix I've added the ` 02` in the location field so that the names are unique and the hack you added is no longer required. 
I still have to think of a proper way to implement this and I'd prefer to not tie the resolver name to the hostname so that users won't have to change their configs if I have to change servers... but that's out of scope here.

TBBle:
Awesome, thank you. I've rebased and rerun the scripts, and the churn is now much lower, so I included the output of the run as a commit for visibility.

v1/dnscrypt-resolvers.csv
```diff
@@ -91,7 +91,10 @@ dnscry.pt-adelaide-ipv4,dnscry.pt-adelaide-ipv4,"DNSCry.pt Adelaide - DNSCrypt,
...
+dnscry.pt-amsterdam-ipv4,dnscry.pt-amsterdam-ipv4,"DNSCry.pt Amsterdam - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,198.140.141.46,2.dnscrypt-cert.dnscry.pt,5A9B:69F3:B181:7B8A:C0E8:18C5:0E97:52A1:D690:C52D:FF92:C8DA:70E4:7551:54E6:19BB,
+dnscry.pt-amsterdam-ipv6,dnscry.pt-amsterdam-ipv6,"DNSCry.pt Amsterdam - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2a03:94e3:222b::1032],2.dnscrypt-cert.dnscry.pt,5A9B:69F3:B181:7B8A:C0E8:18C5:0E97:52A1:D690:C52D:FF92:C8DA:70E4:7551:54E6:19BB,
+dnscry.pt-amsterdam02-ipv4,dnscry.pt-amsterdam02-ipv4,"DNSCry.pt Amsterdam 02 - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,45.86.162.110,2.dnscrypt-cert.dnscry.pt,6E5C:573C:9A33:687D:DDD1:3F61:FFAF:4EA7:59E7:6106:1B5B:8C88:59DF:32A4:E391:9CAF,
+dnscry.pt-amsterdam02-ipv6,dnscry.pt-amsterdam02-ipv6,"DNSCry.pt Amsterdam 02 - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2a07:efc0:1001:a5ce::b4b4],2.dnscrypt-cert.dnscry.pt,6E5C:573C:9A33:687D:DDD1:3F61:FFAF:4EA7:59E7:6106:1B5B:8C88:59DF:32A4:E391:9CAF,
```
TBBle, line 97: Key `6E5C:573C:9A33:687D:DDD1:3F61:FFAF:4EA7:59E7:6106:1B5B:8C88:59DF:32A4:E391:9CAF` is renamed from `dnscry.pt-amsterdam-ipv4` to `dnscry.pt-amsterdam02-ipv4`.

v1/dnscrypt-resolvers.csv
```diff
@@ -141,8 +149,10 @@ dnscry.pt-helsinki-ipv4,dnscry.pt-helsinki-ipv4,"DNSCry.pt Helsinki - DNSCrypt,
...
+dnscry.pt-hongkong-ipv4,dnscry.pt-hongkong-ipv4,"DNSCry.pt Hong Kong - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,89.213.0.26,2.dnscrypt-cert.dnscry.pt,32DD:EDC4:D4D9:B251:982A:373B:D3D9:E117:8929:56DB:8FD9:4FD4:A7AB:1A67:F62D:EF35,
+dnscry.pt-hongkong-ipv6,dnscry.pt-hongkong-ipv6,"DNSCry.pt Hong Kong - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2a13:82c1:850a::b7],2.dnscrypt-cert.dnscry.pt,32DD:EDC4:D4D9:B251:982A:373B:D3D9:E117:8929:56DB:8FD9:4FD4:A7AB:1A67:F62D:EF35,
+dnscry.pt-hongkong02-ipv4,dnscry.pt-hongkong02-ipv4,"DNSCry.pt Hong Kong 02 - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,45.123.188.129,2.dnscrypt-cert.dnscry.pt,0B71:21FC:C3CB:4775:D462:C91E:BEC7:3E47:6D3F:019B:8A69:8D50:CD20:51AD:207E:31C8,
+dnscry.pt-hongkong02-ipv6,dnscry.pt-hongkong02-ipv6,"DNSCry.pt Hong Kong 02 - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2406:4300:bae:6b08::1],2.dnscrypt-cert.dnscry.pt,0B71:21FC:C3CB:4775:D462:C91E:BEC7:3E47:6D3F:019B:8A69:8D50:CD20:51AD:207E:31C8,
```
TBBle, line 155: Key `0B71:21FC:C3CB:4775:D462:C91E:BEC7:3E47:6D3F:019B:8A69:8D50:CD20:51AD:207E:31C8` is renamed from `dnscry.pt-hongkong-ipv{4,6}` to `dnscry.pt-hongkong02-ipv{4,6}`.

v1/dnscrypt-resolvers.csv
```diff
@@ -163,16 +173,20 @@ dnscry.pt-libertylake-ipv4,dnscry.pt-libertylake-ipv4,"DNSCry.pt Liberty Lake -
...
 dnscry.pt-lima-ipv4,dnscry.pt-lima-ipv4,"DNSCry.pt Lima - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,161.132.47.185,2.dnscrypt-cert.dnscry.pt,1CEF:433E:24CB:7D15:68D2:FF63:DD75:B8BC:8FFA:4D40:8091:87E6:BC97:B8C9:4B3F:0775,
 dnscry.pt-lima-ipv6,dnscry.pt-lima-ipv6,"DNSCry.pt Lima - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2001:470:1f2a:1de::2],2.dnscrypt-cert.dnscry.pt,1CEF:433E:24CB:7D15:68D2:FF63:DD75:B8BC:8FFA:4D40:8091:87E6:BC97:B8C9:4B3F:0775,
-dnscry.pt-london-ipv4,dnscry.pt-london-ipv4,"DNSCry.pt London - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,178.239.174.244,2.dnscrypt-cert.dnscry.pt,8F66:DC44:BEBB:62C6:0CEA:2D99:2B92:5FFE:1CBE:FE09:ABB6:6140:8417:6BDB:F2CA:31EE,
+dnscry.pt-london-ipv4,dnscry.pt-london-ipv4,"DNSCry.pt London - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,45.67.84.132,2.dnscrypt-cert.dnscry.pt,8F66:DC44:BEBB:62C6:0CEA:2D99:2B92:5FFE:1CBE:FE09:ABB6:6140:8417:6BDB:F2CA:31EE,
```
TBBle, line 176: IP address was out-of-date, but the host key is the same. IP address manually verified at https://www.dnscry.pt/public-resolvers/lon01

v1/dnscrypt-resolvers.csv
```diff
@@ -163,16 +173,20 @@ dnscry.pt-libertylake-ipv4,dnscry.pt-libertylake-ipv4,"DNSCry.pt Liberty Lake -
...
+dnscry.pt-losangeles-ipv4,dnscry.pt-losangeles-ipv4,"DNSCry.pt Los Angeles - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,84.33.244.100,2.dnscrypt-cert.dnscry.pt,0081:FF5E:2AAB:77C9:4AEC:1980:E72C:16A4:2A14:C835:2746:A518:F03C:71BF:7143:2716,
+dnscry.pt-losangeles-ipv6,dnscry.pt-losangeles-ipv6,"DNSCry.pt Los Angeles - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2a0c:8fc3:3:1:2:3:4:5],2.dnscrypt-cert.dnscry.pt,0081:FF5E:2AAB:77C9:4AEC:1980:E72C:16A4:2A14:C835:2746:A518:F03C:71BF:7143:2716,
+dnscry.pt-losangeles02-ipv4,dnscry.pt-losangeles02-ipv4,"DNSCry.pt Los Angeles 02 - DNSCrypt, no filter, no logs, DNSSEC support (IPv4 server)",,,,2,yes,yes,no,104.200.67.194,2.dnscrypt-cert.dnscry.pt,8871:792B:8640:7C1C:8597:6CB5:0A9C:A0A0:FF44:0B95:E30F:10AF:FD57:9971:59B9:C184,
+dnscry.pt-losangeles02-ipv6,dnscry.pt-losangeles02-ipv6,"DNSCry.pt Los Angeles 02 - DNSCrypt, no filter, no logs, DNSSEC support (IPv6 server)",,,,2,yes,yes,no,[2602:ff75:7:b79::b4b4],2.dnscrypt-cert.dnscry.pt,8871:792B:8640:7C1C:8597:6CB5:0A9C:A0A0:FF44:0B95:E30F:10AF:FD57:9971:59B9:C184,
```
TBBle, line 181: Key `8871:792B:8640:7C1C:8597:6CB5:0A9C:A0A0:FF44:0B95:E30F:10AF:FD57:9971:59B9:C184` is renamed from `dnscry.pt-losangeles-ipv{4,6}` to `dnscry.pt-losangeles02-ipv{4,6}`.

TBBle (commented):
For easier review, I've annotated the changes in the data files that weren't purely new entries. On entry's IP was updated, and three have changed names with a new entry taking their former name, so this should not break anyone's systems unless certificate pinning is common for users in this ecosystem?

Brueggus:
Just a heads-up - the IP addresses of `dnscry.pt-hongkong-ipv4` and `dnscry.pt-hongkong-ipv6` have changed recently. I don't think I can push any changes to this PR.

If we can get this merged, this would help me a lot to keep this repo in sync with changes on my end.

Brueggus:
There's no difference. The JSON is an export of the data my scripts use to
generate the resolver files.

Frank Denis ***@***.***> schrieb am Mi., 18. Sept. 2024,
18:30:

TBBle:
Rebased for latest changes to master branch, which see has had (manual, I assume) updates to the dnscry.pt servers, so the final commit's diff is much smaller now.

> Is there a difference between this and manually parsing the JSON file?

It's not manual. ^_^ For me, that's a win by itself; YMMV.

For example, looking at the current diff, it's readded the three servers you removed as not working in de9d69e32f76f94b57387da45644fa920f0bb57f. (Someone should probably report that to @Brueggus if not already aware...)

It also adds a bunch of anonymous relays which aren't currently in the list. I haven't checked carefully but it looks like that's the list of relays removed in 5762a73773db2d9ab66517c5fe21a3b7d3d0c1ca.

Either way, it makes it super-easy to see what's changed compared to manually parsing a JSON file. (Although I suspect we want to drop the last commit as it's now somewhat reverting deliberate manual changes.)

jedisct1:
But why use the JSON file instead of the already existing `.md` files?

Here are the scripts that have been used to update the `dnscry.pt` entries so far:
- https://github.com/DNSCrypt/dnscrypt-resolvers/blob/next/utils/dnscry.pt-merge.py
- https://github.com/DNSCrypt/dnscrypt-resolvers/blob/next/utils/dnscry.pt-relays-merge.py

They're very simple as they just add a prefix to the names. Using the JSON file looks way more complicated.

TBBle:
> But why use the JSON file instead of the already existing `.md` files?

There's a built-in JSON parser in Python stdlib, so I didn't even need to think about parsing Markdown's various flavours.

I also assumed upstream JSON was canonical, and any MD output was generated from that and liable to change.

When I came into this, both formats already existed and there wasn't any hint of an existing MD parser in the lInked bug or repo that I saw.

Issue #948, opened 2024-08-27
parental-control.md trackers and documentation


bigbgh:
A few of the DNS providers in parental-control.md mention that they block trackers. I have excluded those using `disabled_server_names`.  However, I am still prevented from opening tracker type pages on occasion such as *.awstrack.me/* These are usually links in emails.

1. I've got the IP of the offending DNS server. How can I convert that back to an SDNS key or server host name so I can block that server?
2. Can we please get updated comments in the parental-control.md file to mention ALL the providers which are also blocking trackers.

PR #951, opened 2024-08-28
Update public-resolvers.md

bigbgh:
additional comments that dnsforfamily and its variants also block trackers


jedisct1:
Closes #948

PR #964, opened 2025-01-06
added 0ms DNS

galpt:
added section `## 0ms-dns-unfiltered`

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -23,6 +23,15 @@ To use that list, add this to the `[sources]` section of your
 --
 
 
+## 0ms-dns-unfiltered
+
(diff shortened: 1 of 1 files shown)


jedisct1:
Running on Google Cloud, sending the DNS traffic to Cloudflare, and without support for anonymization :(

We should at least add a certificate hash (`6e62a0efd60a04d93f2f864b54442717a666c038c8c70546af6319e4f46a2f73`).

jedisct1:
[Frontends to other public services (ex: using Cloudflare or Google as a resolver) are not allowed. Beyond collecting additional data, their value is unclear. To hide client IP addresses, running proper DNS relays is recommended instead.](https://github.com/DNSCrypt/dnscrypt-resolvers/blob/master/CONTRIBUTING.md#submitting-update-and-new-servers)

PR #965, opened 2025-01-16
update IP addresses for dnscry.pt-johor-ipv[46]

Brueggus:
IP addresses for jhb01.dnscry.pt have changed.

_I've made the changes only to the `v3` directory as stated in https://github.com/DNSCrypt/dnscrypt-resolvers/blob/master/CONTRIBUTING.md. Please let me know if changes should be added to v1/v2 as well._


jedisct1:
Thank you!

PR #971, opened 2025-02-04
Various improvements to servers I use

a1346054:
Commit messages contain further info.


jedisct1:
Thanks!

I think host names in SNIs were changed to IP addresses in order to help bypass filters. As long as they work, reverting this would not be an improvement.

jedisct1:
The host name is required when other services are hosted on the same IP address, or when it changes the filtering behavior.

jedisct1:
> I have access to a few networks that specifically try to block DoH (and DoT, and redirect traditional DNS too) 

That's why DNS relays exist.

jedisct1:
> Using IP addresses in SNI comes at the cost of clarity, and deviates from how web browsers and other tools use

This is just a string meant to be copied and pasted, or with tools like dnscrypt-proxy, it doesn't even need to be entered.

What really matters is in practice, what advantages do IP addresses have over host names. If one option works more often than another, even just a little bit, it's a better option.

@demarcush made these changes specifically to help avoiding SNI inspection. But I don't know if it really helps or not in practice.

Issue #984, opened 2025-06-06
Add DNS4EU


Ra2-IFV:
https://www.joindns4.eu/for-public#resolver-options

They have:
- Protective resolution
- Protective resolution with child protection
- Protective resolution ad-blocking
- Protective resolution with child protection & ad-blocking
- Unfiltered resolution

And all have both IPv4 and IPv6.
I'm preparing for PR, but isn't that too much? 5 new servers?

Ra2-IFV:
And names, what to do with names?
`dns4eu-protective-child-ad-ipv4` is too long...

Ra2-IFV:
@jedisct1 Do I only need to edit `v3` ?

jedisct1:
Yes! Other directories are updated automatically.

PR #985, opened 2025-06-12
Feat: Add dns4eu

Ra2-IFV:
Close: #984

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -1340,6 +1340,114 @@ Homepage: https://dns4all.eu
 sdns://AgcAAAAAAAAAD1syMDAxOjY3ODo4OjozXQAOZG9xLmRuczRhbGwuZXUKL2Rucy1xdWVyeQ
 sdns://AgcAAAAAAAAAEFsyMDAxOjY3ODo4Ojo2NF0AEGRvcTY0LmRuczRhbGwuZXUKL2Rucy1xdWVyeQ
 
+## dns4eu-ipv4
+
+DNS4EU is an initiative by the European Commission that aims to offer an alternative to the public DNS resolvers currently dominating the market.
+
+Blocks access to known malicious and fraudulent websites. (IPv4 server)
+DNSSEC, No logging.
+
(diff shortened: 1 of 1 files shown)


jedisct1:
Wondering if we shouldn't mention the fact that its name is a bit mistleading: https://techlog.jenslink.net/posts/dns4eu/

Ra2-IFV:
Going to laugh my ass out... I initiated this idea because I read tech new about "Digital sovereignty for Europe" and they mentioned "DNS4EU".
I know the concerns, but it's too late to be an island on today's internet. Either suffer the instant pain or worry about them later. So far Cloudflare is always doing best at what they do. Good reputation. And if the main DNS service is located in Europe it should be fine.
"dns4eu" is their chosen name, it's not wise to change it. Do you have any suggestions for descriptions?

jedisct1:
Certificates are invalid.

Ra2-IFV:
do they rotate everyday? WTF! 
I selected the lowest level of the certificate chain, did I? Then this shouldn't happening!

Ra2-IFV:
Hash check should be skipped: Mine instance resolves to `CN=resolver.joindns4.eu` and your is `CN=hos.joindns4.eu`, they use LetsEncrypt E5 and E6 certificates

Ra2-IFV:
@jedisct1 PR check awaiting approval

Ra2-IFV:
**NOTICE**

They have 2 IPs for each DNS service. The `IP Address` field in the stamp calculator didn't say it can be comma-seperated. What to do now?

Ra2-IFV:
```text
[2025-09-19 00:50:17] [NOTICE] -   30ms dns4eu-unfiltered-ipv4
```

Of course it works since now we use bootstrap servers to resolve hostname. But this reduces stability! Removing hard-coded IP address in the stamp may mess up IPv4 and IPv6 since it's all depends on the bootstrap server's anwser.

Give the order, sir! 20 entries (2 servers for IPv4 and IPv6 each) for 5 DNS services, or 5 for 5?

PR #997, opened 2025-09-13
Propose odoh-snowstorm as public odoh target

willscott:


jedisct1:
Unfortunately, that doesn't work:

```text
[2025-09-16 18:00:47] [WARNING] Unable to decrypt response from [odoh-snowstorm]: [Malformed response]
```

jedisct1:
Lack of padding is an issue, as DoH/ODoH don't do any length hiding at transport layer.

> When trying it against odohrelay-crypto-sx I get Did not obtain the correct headers from dope.snowstorm.love:443 with response Will only relay to known public ODoH server. - is this expected?

Seems to work fine here. Are you using a custom stamp? Or just `servers = ['oooh-snowstorm']` ?

Issue #442, opened 2021-03-12
CIRA dns servers


skhan85:
Hi why only CIRA-Family dns is available with DNSCRYPT. CIRA has two other servers too: CIRA Private and CIRA Protect. But using e.g. CIRA-Protect in the .toml file doesn’t work. Using Raspberry PI. 

https://www.cira.ca/node/8331

jedisct1:
CIRA servers are very buggy.

They are the only ones that only support a minimal set of query parameters, and they if the EDNS-clientsubnet option is used, they are unable to process the query and return a `REFUSED` response.

So, if they are in the list like other servers, queries will randomly fail when CIRA is being chosen.

It would be good to have them back. Maybe once these bugs are finally fixed.

PR #1007, opened 2025-12-10
Add CIRA Canadian Shield DNS resolver (DoH)

lexcyn:
- Added the 3 CIRA Canadian Shield DNS resolvers (Private, Protected & Family) with IPv6 options
- https://www.cira.ca/en/canadian-shield/configure/summary-cira-canadian-shield-dns-resolver-addresses/


lexcyn:
> CIRA was removed 4 years ago: [#442 (comment)](https://github.com/DNSCrypt/dnscrypt-resolvers/issues/442#issuecomment-797488366)
> 
> Did they manage to fix their servers since?

I think so - I've been using it via the cloudflared dns-proxy for the past year with no issues!

lexcyn:
> In the description, can you clarify what "private" and "protected" mean?
> 
> It's basic DoH without anonymization, so "private" may be confusing.

Updated the descriptions. Let me know if that is sufficient

jedisct1:
See https://github.com/DNSCrypt/doh-server?tab=readme-ov-file#dns-stamps-and-certificate-hashes

But not a big deal, I will add them.

jedisct1:
Added: https://status.dnscrypt.info

Thank you!

PR #1017, opened 2026-02-16
Update Safe Surfer servers

kirwan-safesurfer:
General update, including DoH primary and secondary IPv4 + IPv6.


kirwan-safesurfer:
Thanks, try it now. I was able to produce the corrected result alongside matching the passing secondary hash.

PR #1024, opened 2026-03-05
Add DNSCry.pt Vienna to relays.md

A-Loot:
Not sure why it was missing.


jedisct1:
Probably because it didn't work :)

PR #1031, opened 2026-04-05
Re-add: saldns02-conoha-ipv4 and anon-saldns02-conoha-ipv4

junkurihara:
Hello!

I found that a DNSCrypt server and a relay (saldns02-conoha-ipv4 and anon-...) have been removed last week. (https://github.com/DNSCrypt/dnscrypt-resolvers/commit/cf121e2b26a68f6956882b61a8b1f39273234e59)

At that time, the VPS provider (ConoHa) faced a large outage due to storage failure for almost 24hours, and the server and relay were quite unstable. Now it is back and working as normal. So please add them again.

Thanks!


jedisct1:
Thank you!

PR #1032, opened 2026-04-06
update dct servers

dct-infra:
- added dct-de2
- changed dct-fr IP - old fr server will stay on for a few days.


jedisct1:
Thank you!

PR #1037, opened 2026-04-23
Update onion-services url

syphyr:
Remove non-functional https://ipv6.download.dnscrypt.info source. Add new https://cdn.jsdelivr.net source.


jedisct1:
Done in 0221744b364985b2e4febf12514ba09b61aeb14b - Thanks!

PR #1045, opened 2026-05-12
Add sorcerousmachine resolver (OpenNIC Tier-2)

sorcmach:
Adds four entries (DNSCrypt v4/v6, DoH v4/v6) for a new public OpenNIC
Tier-2 resolver: dns.sorcerousmachine.com.

- Listed at servers.opennic.org as a public Tier-2.
- No-logs, non-filtering.
- DNSSEC validation against IANA + OpenNIC root anchors.
- Stamps validated locally with dnscrypt-proxy 2.x prior to submission.
- Names appended to v3/opennic.subset for inclusion in opennic.md.

Note on DoH cert hashes: intentionally omitted. The resolver uses
Let's Encrypt with automatic ~60-day renewal, so a pinned hash would
invalidate the registry entry on every renewal. Clients fall back to
system trust roots, same as a browser.

### DIFF
--- a/v3/opennic.subset
+++ b/v3/opennic.subset
@@ -30,3 +30,7 @@ opennameserver-4-doh
 opennic-R4SAS
 publicarray-au-doh
 publicarray-au2-doh
+sorcerousmachine
+sorcerousmachine-doh
+sorcerousmachine-doh-ipv6
+sorcerousmachine-ipv6
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -5058,6 +5058,38 @@ Maintained by Frank Denis - https://fr.dnscrypt.info/sfw.html
 sdns://AQMAAAAAAAAADzE2My4xNzIuMTgwLjEyNSDfYnO_x1IZKotaObwMhaw_-WRF1zZE9mJygl01WPGh_x8yLmRuc2NyeXB0LWNlcnQuc2Z3LnNjYWxld2F5LWZy
 
 
+## sorcerousmachine
+
+Public, no-logs, non-filtering OpenNIC Tier-2 recursive resolver. DNSSEC against IANA and OpenNIC root anchors. Operated by Sorcerous Machine, LLC in Warrenton, VA, US.
+https://dns.sorcerousmachine.com/
+
+sdns://AQcAAAAAAAAAEjE1LjIwNC4xMzUuMjg6ODQ0MyBMpSYYYGo1AnEYIVggIK6tOA-aNBsMnVpbm8QvfUn23ygyLmRuc2NyeXB0LWNlcnQuZG5zLnNvcmNlcm91c21hY2hpbmUuY29t
+
+
+## sorcerousmachine-doh
+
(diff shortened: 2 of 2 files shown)


jedisct1:
Hi,

And thanks for these resolvers!

Looks like the DNSSEC flag was set but the servers don't support DNSSEC: https://github.com/DNSCrypt/dnscrypt-resolvers/actions/runs/25706167815/job/75516339536?pr=1045


Regarding hashes: what usually gets pinned is the CA (which changes very infrequently), not leaves.

sorcmach:
Thanks for the careful review.

You were right. The server-side DNSSEC check was failing. Root cause: this resolver was set up in the canonical OpenNIC Tier-2 layout that slaves the OpenNIC root zone alongside the 16 TLDs. That makes BIND authoritative for the root, so root-level NXDOMAIN comes back with the `AA` flag, and per [RFC 6840 §5.8](https://datatracker.ietf.org/doc/html/rfc6840#section-5.8) authoritative responses must not carry `AD`. `dnscrypt-proxy`'s self-check probes `*.dnscrypt-test.` and expects `AD` on NXDOMAIN, which the slaved-root layout architecturally can't provide.

Fixed by switching to a hybrid layout: IANA root hints for recursion, the 16 OpenNIC TLDs still served from local slaves, and a small forward zone for `glue.` covering OpenNIC infrastructure names. ICANN names now recurse through the IANA hierarchy and produce real `AD`-flagged DNSSEC end-to-end, including NXDOMAIN.

Local verification against the production stamp:

```
$ dnscrypt-proxy -config check.toml   # require_dnssec = true
[NOTICE] [sorcerousmachine-dnscrypt] OK (DNSCrypt) - rtt: 0ms
[NOTICE] Server with the lowest initial latency: sorcerousmachine-dnscrypt (rtt: 0ms), live servers: 1
```

Could you re-run CI when convenient? Happy to push a commit if that's easier than re-running from your side.

sorcmach:
Withdrawing this PR for now.

To make `dnscrypt-proxy`'s `require_dnssec` check pass, our resolver had to switch from slaving the OpenNIC root to using IANA root hints with a forward zone for `.glue`. That fix worked for the DNSSEC check (root-level NXDOMAIN comes back with `AD` from the IANA hierarchy), but it broke OpenNIC's own hourly health check, which probes `. SOA / . NS / . DNSKEY` against every listed Tier-2 and expects authoritative OpenNIC root data. The resolver showed Offline in the OpenNIC member portal for ~48 hours straight and was at imminent risk of being auto-delisted from the OpenNIC servers page.

The two checks ask for incompatible behaviors at the root: dnscrypt-proxy wants recursive `AD` on NXDOMAIN, OpenNIC wants authoritative `AA` from a slaved root, and RFC 6840 §5.8 forbids setting `AD` on `AA` responses. As an OpenNIC Tier-2, OpenNIC listing has to win.

The resolver is operational and DNSSEC-validates ICANN names correctly; it just can't satisfy dnscrypt-proxy's specific synthetic `*.dnscrypt-test.` probe while remaining an OpenNIC Tier-2. I'll re-submit if a workaround emerges (response rewriting, an upstream change to dnscrypt-proxy's probe, or a different way of expressing DNSSEC capability in the stamp).

Thanks for the careful review.

PR #1047, opened 2026-05-18
Add vpsdns.co.za public DNS services information

rb-sys:
Added public DNS services in South Africa with details on DoH, DoT, and DoQ features.


jedisct1:
Thanks!

These list don't include DoT and DoQ.

Also, could you add hashes for the DoH entry? See https://github.com/DNSCrypt/dnscrypt-resolvers/blob/master/CONTRIBUTING.md#doh-servers

jedisct1:
The stamp is invalid. Make sure you are actually using this server. If not, it will be difficult to maintain it :) 

The frontend is Cloudflare, which is worth mentioning as people may be concerned about this and the fact that it may contradict the strict no-log policy.

PR #1048, opened 2026-05-23
Rb sys patch 1

rb-sys:
Adds two new South African public encrypted DNS resolvers operated on dedicated VPS infrastructure.

DoH now resolves to the internal VPS IP i.e Grey Cloud, to adhere to the no-logs policy. 

The domain is utilizing a Let's Encrypt ECDSA certificate. The cryptographic stamp has been completely recalculated and updated using the exact SHA-256 SPKI public key pin (`e6f11ec806c1c7e64c2830acbaf49a8993550bba445af565fb2127716f1d2248`) advertised by the live TLS handshake. Certbot will also renew the cert with reuse_key = True, ensuring that the SPKI key won't change.

Please let me know if there is anything else you're concerned about/wrong.


jedisct1 (changes requested):
Thanks a lot for the PR, and for iterating on the stamps!

I checked the three new stamps locally:

- `vpsdns-za` DNSCrypt: resolves successfully
- `vpsdns-za-doh` DoH: resolves successfully
- `anon-vpsdns-za` relay: works as an anonymized DNSCrypt relay

One formatting issue needs to be fixed before this can be merged: v3 entries must use the repository's plain entry format, with the description as plain text and each stamp on its own line starting directly with `sdns://`. Lines such as `Name:`, `DNSCrypt: sdns://...`, `DoH: sdns://...`, `Stamp: sdns://...`, `IPv6:`, `DNSSEC:`, etc. are treated as description text by `utils/format.py`, so the new entries currently have no parseable stamps and would be dropped/flagged as invalid by the formatter.

Please change the entries to something like:

```md
## vpsdns-za

High-performance privacy DNSCrypt resolver based in South Africa.

sdns://AQcAAAAAAAAAEjEwMi4yMTQuMTAuODI6ODQ0MyAp_ZK8Ab77yIXFI7AIeSrgjZjUJ2zG9acKC0XARJZprRwyLmRuc2NyeXB0LWNlcnQudnBzZG5zLmNvLnph

## vpsdns-za-doh

High-performance privacy DoH resolver based in South Africa.

sdns://AgcAAAAAAAAADTEwMi4yMTQuMTAuODIg5vEeyAbBx-ZMKDCsuvSaiZNVC7pEWvVl-yEncW8dIkgQZG9oLnZwc2Rucy5jby56YQovZG5zLXF1ZXJ5
```

And for the relay:

```md
## anon-vpsdns-za

Anonymized DNSCrypt relay based in South Africa.

sdns://gRIxMDIuMjE0LjEwLjgyOjg0NDM
```

After that, please run the formatter (or just push the change and I can re-check). The actual stamps look good from here. Thanks again!

rb-sys:
Updated the blocks to match the plain entry format and stripped out the configuration attributes. Ready for re-check!

jedisct1 (approved):
Thanks for the quick update! The entries now use the expected plain v3 format, and I re-checked the current stamps locally:

- `vpsdns-za` DNSCrypt: OK
- `vpsdns-za-doh` DoH: OK
- `anon-vpsdns-za` relay: OK

Looks good from my side. Thanks again for contributing these South African resolver/relay entries!

jedisct1:
Thanks again for the contribution!

One small issue showed up in CI: the DoH stamp advertised DNSSEC support, but when tested through dnscrypt-proxy/dnslookup the resolver reported `DNSSEC: no, the resolver doesn't support DNSSEC`. Because clients use the stamp properties to know what capabilities to expect, advertising DNSSEC when it is not detected makes the check fail.

I removed only the DNSSEC flag from the DoH stamp and adjusted that DoH description accordingly. The resolver itself still works fine, and the DNSCrypt stamp continues to advertise DNSSEC support.

rb-sys:
Hi,

Thank you for allowing me to be a part of this project!

I tested the resolver and noticed something interesting regarding DNSSEC detection.

Even though the DoH stamp no longer advertises DNSSEC support, the resolver is still fully validating DNSSEC in practice:

- dnssec-failed.org correctly returns SERVFAIL
- invalid DNSSEC chains are being rejected
- dig +dnssec responses show validation enforcement (EDE codes present)

So DNSSEC is functionally active at the resolver level.

However, dnscrypt-proxy’s capability detection still reports “DNSSEC: no”, likely because it relies on stamp metadata rather than live validation probing behavior.

This creates a mismatch where:
✔ DNSSEC is operational
❌ DNSSEC capability is not detected from the stamp

Just wanted to confirm whether this is intentional (strict metadata policy) or whether there’s any preferred way to signal DNSSEC support for DoH resolvers that validate but currently have the capability flag omitted from their deployment stamp.

Happy to provide logs if useful.

jedisct1:
Hi! Thanks for checking this more deeply — and yes, your observations are useful.

`dnssec-failed.org` returning `SERVFAIL` is a good sign: it means the resolver is rejecting bogus signatures. However, the DNSSEC flag in DNS stamps has to mean a bit more than that for clients: the resolver must also signal authenticated data correctly, including authenticated denial of existence.

I re-tested the DoH endpoint with the DNSSEC bit enabled in the stamp. The live test reports:

```text
DNSSEC        : incomplete, the resolver rejects bogus signatures but doesn't validate authenticated denial of existence
```

The important detail is the negative-answer test. A query for `nonexistent-zone.dnscrypt-test` returns `NXDOMAIN` with DNSSEC records, but the response flags are:

```text
flags: qr aa rd ra
```

There is no `ad` flag. So from a client perspective, the resolver is not saying “this negative answer has been authenticated”. That is why dnscrypt-proxy does not consider the DoH endpoint as fully DNSSEC-capable, even though bogus signatures are rejected.

So this is not just dnscrypt-proxy reading the stamp metadata. The stamp says what the resolver advertises, then the resolver is probed live. If the stamp advertises DNSSEC, the live DoH behavior has to match, otherwise clients may make incorrect assumptions.

For now, I think the safest thing is to keep DNSSEC advertised on the DNSCrypt stamp, but not on the DoH stamp until the DoH service sets/propagates the AD bit correctly for validated answers, including authenticated denials.

If you want a DoH implementation that is known to work well for this use case, I’d recommend https://github.com/dnscrypt/doh-server. It was built for this ecosystem, can be placed in front of a validating resolver, and also supports Oblivious DoH, which would be a nice future option for this service.

Thanks again for contributing this — the resolver is useful, this is just an important DNSSEC signaling detail we need to get right before advertising the DNSSEC property on the DoH stamp.

rb-sys:
Hi,

Thank you so much for the deep dive and the live validation analysis! That makes perfect sense. 

I see exactly what you mean now regarding the missing AD flag on the negative-answer test (NXDOMAIN).

Thank you also for pointing me toward the dnscrypt/doh-server project! I’m going to look into deploying that as a frontend proxy layer to handle the header adjustments and potentially unlock Oblivious DoH down the line.

Appreciate your guidance on getting these signaling details right for the ecosystem. Glad to be a part of the project!

PR #1052, opened 2026-05-28
CipherDNS endpoint addition ( formally vpsdns.co.za )

rb-sys:
Hi there. 

As you may have noticed, I've migrated from the vpsdns.co.za domain towards cipherdns.co.za, hence why the endpoints stopped working. Apologies for any inconvenience.

Thank you !


jedisct1:
Thanks a lot for the update, Rouve! The migration to the CipherDNS domain makes sense, and the new DNSCrypt, DoH, and anonymized relay stamps all check out. Happy to have these South Africa endpoints back in the list. 🙌

jedisct1 (approved):
Approved — thanks again for updating the endpoints and relay details for CipherDNS.

PR #1057, opened 2026-06-19
feat: more backup urls

fir4tozden:
More backups have been added and beautified.


jedisct1:
Although the files are markdown for the markup conventions, software rendering this is not expected to render markdown. So, sections like "```toml " will render as-is.

I'm not sure that adding more download URLs is very useful, especially since there are 3 CDNs already.

PR #1066, opened 2026-07-10
DNSCry.pt DOH Servers

ghost:
DNSCry.pt DOH Servers


jedisct1:
Thanks!

As they are frequently updated, the `dnscry.pt` servers and relays are automatically updated from lists they publish at URLs like https://www.dnscry.pt/resolvers.md

The DoH list is going to be added. Thanks!

jedisct1:
More intersting: they support ODoH :)

PR #1071, opened 2026-07-17
New format suggestion

ghost:
What do you think about this format for each listings?

```

## adfilter-adl-ipv6

| Name |   |
| -- | -- |
| Operator | ADFilter |
| Homepage | https://adfilter.com/ |   <-------------- if not exist, "-".
| Type |  DOH |   <-------------- or DNSCrypt
| Comment | Hosted in Adelaide, Australia. No EDNS Client-Subnet. |
| Has DNSSEC? | ✔️ |
| No logging? | ✔️ |
| No Filter? | ❌ Blocks ads, malware, trackers and more. |   <-------------- if "X", write more details like this

`sdns://AgMAAAAAAAAAGFsyNDA0Ojk0MDA6NjkwMzoyMTAwOjoxXSCQeNIaVsDgiR2QYiG2mSMxvdFxcusPM92gfVxlxuFMORBhZGwuYWRmaWx0ZXIubmV0Ci9kbnMtcXVlcnk`

```

This is just an example.

### DIFF
--- a/v3/public-resolvers.md
+++ b/v3/public-resolvers.md
@@ -25,94 +25,141 @@ To use that list, add this to the `[sources]` section of your
 
 ## a-and-a
 
-Andrews & Arnold non-filtering resolver.
-No logging, DNSSEC validation. Homepage: https://www.aa.net.uk/dns/
-Operated by Andrews & Arnold. Service page: https://www.aa.net.uk/dns/
+| Name |   |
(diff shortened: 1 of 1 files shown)


jedisct1:
Flags are already present in the stamp. 

And we can't break all the applications consuming these lists for no reasons.

PR #1075, opened 2026-08-04
Mullvad is planning to deprecate their DoH provider

ghost:
https://github.com/mullvad/mullvad-browser/issues/577


jedisct1:
Thank you!

iyanmv:
This was completely early and unnecessary...

iyanmv:
All my servers lost connectivity because of this change, completely unannounced from your side, and honestly, for no good reason, since all the Mullvad DoH servers are still usable.

ghost:
> All my servers lost connectivity because of this change

You're relying on only 1 provider? That's your problem LOL

iyanmv:
> > All my servers lost connectivity because of this change
> 
> You're relying on only 1 provider? That's your problem LOL

Instead of laughing at setups you don't understand and blaming something on me, I'd recommend you are more careful with your contributions in the future. You broke this, not Mullvad. And my setup is just fine for what I want/expect.

I've been using Mullvad DNS servers for years and never had more than few minutes of downtime. They are that reliable that I can configure my resolvers with just one provider, and without a fallback. If it fails, I want to investigate and understand why, rather than using the fallback without noticing a thing. I've been using Mullvad DNS servers with dnscrypt since I updated this https://github.com/DNSCrypt/dnscrypt-resolvers/pull/827 and before that with cloudflared. So I think my expectations are not that far off. You managed to break this for me (and others) with a change that no one asked for, and was not needed.

I hope based on your reply here, your contributions in this project are double and triple checked before you break more things in the future.

PR #1084, opened 2026-09-15
Add Dremaxx resolvers

yannickdreher:
Hi,

this adds the DoH resolvers I run at Dremaxx (https://dremaxx.de/en/dns-resolvers). They are recursive on our own servers in Zurich, Dusseldorf and Houston, IPv6 is anycast. The family resolvers block adult content, so I also added them to the parental-control list.

```
Entry                 Addresses                                      Props
dremaxx               185.44.82.158, 194.28.99.248, 185.121.169.47   DNSSEC, No-log, No-filter
dremaxx-ipv6          2a0c:9a40:e100::53, 2a0c:9a40:e101::53         DNSSEC, No-log, No-filter
dremaxx-family        185.44.82.150, 62.3.50.42, 185.121.169.45      DNSSEC, No-log
dremaxx-family-ipv6   2a0c:9a40:e100::f53, 2a0c:9a40:e101::f53       DNSSEC, No-log
```

The stamps include the certificate hash. I tested all of them with dnscrypt-proxy, with and without HTTP/3.

Thanks!


jedisct1:
Thanks for this.

I tested all four entries from my side. They all work. The cert hash is the right one, the addresses match DNS, and both resolvers answer over HTTP/2 and HTTP/3.

The blocker is CI. `prcheck` fails on this commit, and it also failed on the first version of this branch.

I compared the two revisions. The version that passed CI had one stamp less, and that is the only difference between them:

```
sdns://AgMAAAAAAAAACjYyLjMuNTAuNDIgjFAzlz_-T_HvCqkphPyTKBunqSeO-L9kEia4mNENTnQWZmFtaWx5LmRucy5kcmVtYXh4LmNvbQovZG5zLXF1ZXJ5
```

That is `family.dns.dremaxx.com` on `62.3.50.42`. Every other stamp is identical in both revisions.

That address is fine from my network, so this looks like a reachability problem from GitHub. The server may be slow for GitHub runners, or it may filter their traffic.

I will re-run `prcheck` once, in case it was a flake. If it fails again, the easiest fix is to drop that stamp. You can add it back later, once your provider confirms that `62.3.50.42` accepts traffic from GitHub.

One small note: `v3/parental-control.md` is generated from `v3/parental-control.subset`. You only need to edit the subset file. I run `make all` before merging, so you don't need to touch the generated files.

Everything else looks good. Thanks!

yannickdreher:
Thanks a lot for testing all four entries and for digging into the CI failure.
That was really helpful.

You were right about 62.3.50.42: that address had an MTU problem, which I've
fixed with MSS clamping, and it passed in the previous prcheck run. That run
failed on a different stamp though, the Zurich one (185.44.82.150). That
server is hosted with iFog GmbH and currently loses about 1% of SYNs on their
side. I've raised it with them, and it would also explain why it passes some
runs and not others.

I've reverted parental-control.md as you suggested and rebased onto current
master. The checks for the new push are waiting for approval. If prcheck
still flakes on the Zurich stamp, I'm happy to drop that one until iFog has
sorted it out.

Thanks again!

jedisct1:
Added! Thank you!

PR #1087, opened 2026-09-23
Add anon-cryptids-relay for IPv4 and IPv6

PabloReinhardt:
Added new DNSCrypt anonymized relay entries for IPv4 and IPv6 endpoints.


jedisct1:
Hi!

And thanks a lot for contributing these relays!

Unfortunately the IPv6 one doesn't seem to be reachable. Can you check?

PabloReinhardt:
> Hi!
> 
> And thanks a lot for contributing these relays!
> 
> Unfortunately the IPv6 one doesn't seem to be reachable. Can you check?

Hi! 

Thanks for the heads up - I found the issue, I had forgotten to add a corresponding ufw rule for TCP/8443, so the port was filtered out. I added the rule now and confirmed from an external machine that TCP/8443 is reachable over IPv6. 

Thank you for pointing this out!

